PR #8, opened 2017-06-27
remove Azure prefix to make class name shorter

ZhijunZhao:
@yungez @xscript @aovechki


azure-storage-spring-boot-autoconfigure/src/main/java/com/microsoft/azure/autoconfigure/azurestorage/StorageProperties.java
```diff
@@ -9,7 +9,7 @@
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 @ConfigurationProperties("azure.storage")
-public class AzureStorageProperties {
+public class StorageProperties {
 
     private String connectionString;
```
rajadilipkolli: Instead of one string it would be better to add three values like DefaultEndpointsProtocol , AccountName and AccountKey and form connectionstring in java.

Any suggestions is welcome. If you are ok with this I will submit PR

ZhijunZhao: Actually, I thought about this solution before I wrote the current implementation. The reason of not using it is that it's more complicated for both implementation and end user usage. 

1. User can simply copy the connection string from Azure portal.
2. All sample code I can find uses the single connection string parameter, instead of several parameters. https://github.com/Azure-Samples/storage-blob-java-getting-started/blob/master/src/BlobClientProvider.java

Unless you can provide a strong reason or a sample, I'll keep the current implementation.

PR #15, opened 2017-06-30
updates snapshot version to follow release version

rajadilipkolli:
updates snapshot version to follow maven release version, we are not following standard version naming . It should be updated automatically. 

If you are Ok I can provide a PR to use maven release plugin which seamlessly allows to update versions and release project.


ZhijunZhao:
Sure, that would be great!

PR #19, opened 2017-07-03
intial checkin of azure documentdb spring data repository support

yungez:
@ZhijunZhao @aovechki  @selvasingh


spring-data-azure-documentdb/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+    <developers>
+        <developer>
+            <id>yungez</id>
+            <name>Yunge Zhu</name>
+            <email>yungez@microsoft.com</email>
+        </developer>
+    </developers>
+
+    <properties>
+        <springdata.commons.version>1.13.4.RELEASE</springdata.commons.version>
```
rajadilipkolli: We should get dependencies from azure-spring-boot-dependencies module and as we are importing as parent POM

spring-data-azure-documentdb/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+        <springcore.version>4.3.9.RELEASE</springcore.version>
+        <mockito.version>1.10.19</mockito.version>
+    </properties>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-core</artifactId>
+            <version>${springcore.version}</version>
```
rajadilipkolli: We are not taking advantage of Spring boot's BOM , We shouldn't declare spring version here and we should get dependencies from azure-spring-boot-dependencies module and as we are importing as parent POM

spring-data-azure-documentdb/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+            <version>1.11.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.0</version>
+        </dependency>
+
+        <dependency>
```
rajadilipkolli: Instead of junit, spring-test and hamcrest-all we should use spring-boot-starter-test dependency

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/Utils.java
```diff
@@ -0,0 +1,12 @@
+package com.microsoft.azure.spring.data.documentdb;
+
+public class Utils {
+
+    public static void printInfo(String msg) {
+        System.out.println(msg);
```
rajadilipkolli: Shouldn't we use log here?

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/config/AbstractDocumentDbConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+import com.microsoft.azure.spring.data.documentdb.core.DocumentDbTemplate;
+import com.microsoft.azure.spring.data.documentdb.core.convert.DocumentDbConverter;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public abstract class AbstractDocumentDbConfiguration extends DocumentDbConfigurationSupport {
+
+    @Value("${azure.documentdb.databasename}")
```
rajadilipkolli: Better to use @ConfigurationProperties as suggested by spring to use in spring 4+ versions , though @Value will work and there is nothing wrong here

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/config/DocumentDbConfigurationSupport.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.data.annotation.Persistent;
+import org.springframework.data.mapping.context.MappingContext;
+import org.springframework.data.mapping.context.MappingContextIsNewStrategyFactory;
+import org.springframework.data.mapping.context.PersistentEntities;
+import org.springframework.data.support.CachingIsNewStrategyFactory;
+import org.springframework.data.support.IsNewStrategyFactory;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.StringUtils;
+
+import java.util.*;
```
rajadilipkolli: From performance standpoint its always better to use static imports instead of util.*

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
```diff
@@ -0,0 +1,221 @@
...
+            }
+            return collection;
+        } catch (DocumentClientException e) {
+            throw new RuntimeException(e.getMessage());
+        }
+
+    }
+
+    private DocumentCollection getCollection(String dbName, String collectionName) {
+        final String query = "SELECT * FROM root r where r.id='" + collectionName + "'";
```
rajadilipkolli: From performance standpoint its better to use Stringbuilder. if we call getCollection 1000 times then 1000 objects will be stored in string pool and wont be available for garbage collection. Instead if we use StringBuilder it will be available for garbage collection

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/repository/DocumentDbRepository.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.data.repository.PagingAndSortingRepository;
+
+import java.io.Serializable;
+import java.util.List;
+
+@NoRepositoryBean
+public interface DocumentDbRepository<T, ID extends Serializable> extends PagingAndSortingRepository<T, ID> {
+
+    @Override
+    List<T> findAll();
```
rajadilipkolli: If we are supporting java 8+ then I feel we should also support Stream

spring-data-azure-documentdb/pom.xml
rajadilipkolli: Declaring versions here should be used to override the version provided by spring-boot BOM

spring-data-azure-documentdb/pom.xml
yungez: thanks, resolved

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/config/DocumentDbConfigurationSupport.java
yungez: it's heavy to import static utils, changed to specific packages

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/repository/DocumentDbRepository.java
yungez: it's not determined on timeline to support Stream yet. not in scope of this initial checkin.

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/Utils.java
yungez: nice. thanks.

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
yungez: that's good point. while given those string is not intern, and no loop or complex case , there's no big difference by using StringBuilder and string concatentation. while string concatentation readability is good. so keep current way. https://stackoverflow.com/questions/4645020/when-to-use-stringbuilder-in-java
https://stackoverflow.com/questions/8725739/correct-way-to-use-stringbuilder

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/repository/DocumentDbRepository.java
rajadilipkolli: Thank you, should I raise issue so that we will not loose track of it

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
rajadilipkolli: thank you, good learning for me too

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/repository/DocumentDbRepository.java
yungez: Sure. it would be nice if you could help add some reasons.

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/DocumentDbFactory.java
```diff
@@ -0,0 +1,29 @@
+/**
```
ZhijunZhao: Shall we move this file to the core folder?

yungez: not necessary. core contains operation/context mapping context. and current folder organization is following spring data mongo, graph database, cassandra examples.

PR #20, opened 2017-07-03
rework pom files

ZhijunZhao:
@yungez @aovechki @xscript


pom.xml
```diff
@@ -4,12 +4,12 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>com.microsoft.azure</groupId>
-    <artifactId>azure-spring-boot-starters</artifactId>
-    <version>0.1.2-SNAPSHOT</version>
+    <artifactId>azure-spring-boot-starter-bom</artifactId>
+    <version>0.1.2</version>
```
rajadilipkolli: Shouldn't the version which is under development be declared as SNAPSHOT? If not can you please let us what is the process Microsoft uses.

ZhijunZhao: We use Nexus Repository Manager. https://oss.sonatype.org/

We have CI setup against it from another team. 

You're right that the version under development should be declared as SNAPSHOT. I tried it but it didn't work as expected. Haven't spent much time looking into the root cause. Will revisit this part Iater.

You can log an issue if you like.

PR #24, opened 2017-07-05
add sample code under spring-data-azure-documentdb-sample

yungez:
@ZhijunZhao  @aovechki


spring-data-azure-documentdb-sample/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>spring-data-azure-documentdb</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>azure-documentdb</artifactId>
+        </dependency>
```
ZhijunZhao: is this dependency needed?

spring-data-azure-documentdb-sample/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+                    <execution>
+                        <goals>
+                            <goal>repackage</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
```
ZhijunZhao: I think it should also work if you remove this plugin

spring-data-azure-documentdb-sample/src/main/java/com/microsoft/azure/sample/SampleApplication.java
```diff
@@ -0,0 +1,32 @@
...
+
+        final User testUser = new User("testId", "testFirstName", "testLastName", "test address line one");
+
+        repository.deleteAll();
+        repository.save(testUser);
+
+        final User result = repository.findOne(testUser.getId());
+
+        Assert.state(result.getFirstName().equals(testUser.getFirstName()), "query result firstName doesn't match!");
+        Assert.state(result.getLastName().equals(testUser.getLastName()), "query result lastName doesn't match!");
```
ZhijunZhao: How about removing the assert and using log `user xxx found`.

spring-data-azure-documentdb-sample/pom.xml
yungez: yes,

spring-data-azure-documentdb-sample/pom.xml
yungez: i'd like to keep it so that the sample could be executed standalone

spring-data-azure-documentdb-sample/src/main/java/com/microsoft/azure/sample/SampleApplication.java
yungez: good point! fixed

PR #27, opened 2017-07-12
optimize documentdb query based on service team feedback

yungez:
@ZhijunZhao @aovechki 

1. avoid query collection, instead of use constructed collection link
2. use readDocument instead of queryDocument for performance consideration

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,6 +26,12 @@
             <name>Zhijun Zhao</name>
             <email>zhijzhao@microsoft.com</email>
         </developer>
+        
+         <developer>
+            <id>yungez</id>
+            <name>Yunge Zhu</name>
+            <email>yungez@microsoft.com</email>
+        </developer>
     </developers>
 
     <scm>
--- a/spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
+++ b/spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
@@ -28,6 +28,9 @@ public class DocumentDbTemplate implements DocumentDbOperations, ApplicationCont
     private final DocumentDbConverter dbConverter;
     private final String databaseName;
 
+    private Database databaseCache;
+    private List<String> collectionCache;
+
     public DocumentDbTemplate(String host,
                               String key,
                               String dbName,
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -26,6 +26,12 @@
             <name>Zhijun Zhao</name>
             <email>zhijzhao@microsoft.com</email>
         </developer>
+        
+         <developer>
```
ZhijunZhao: reformat? it seems the tab/space is different.

yungez: it's reformatted already, all space.

yungez: it's reformatted already, all space

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
```diff
@@ -285,6 +286,10 @@ private String getDatabaseLink(String databaseName) {
     }
 
     private String getCollectionLink(String databaseName, String collectionName) {
```
ZhijunZhao: Now that database name is a data member, I prefer to remove this parameter.

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
```diff
@@ -43,71 +46,145 @@ public DocumentDbTemplate(DocumentDbFactory documentDbFactory,
...
-    public <T> void delete(T objectToRemove) {
-        throw new UnsupportedOperationException("not supported");
+    public String getCollectionName(Class<?> entityClass) {
+
+        return entityClass.getSimpleName();
     }
 
-    private Database getDb(String dbName) {
+    private Database createOrGetDatabase(String dbName) {
         final String query = "SELECT * FROM root r WHERE r.id='" + dbName + "'";
```
ZhijunZhao: I think below code section 

                if (resource instanceof Database) {
                    db = (Database) resource;
                }
                return db;

should be changed to 

                if (resource instanceof Database) {
                    return  (Database) resource;
                } else {
                     return null;
                }

Although origin code works, the logic is not correct.

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
yungez: interface should be general. like collectionName, although could get from entityClass, but still use have choice to customize it

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
yungez: nice catch, fixed

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
```diff
@@ -158,32 +234,27 @@ public DocumentCollection createCollection(String collectionName, RequestOptions
...
+        final List<DocumentCollection> collectionList = documentDbFactory.getDocumentClient()
+                .queryCollections(getDatabaseLink(dbName), query, null).getQueryIterable().toList();
 
-                final RequestOptions requestOptions = new RequestOptions();
-                requestOptions.setOfferThroughput(1000);
+        if (collectionList.size() > 0) {
+            return collectionList.get(0);
+        } else {
+            DocumentCollection collection = new DocumentCollection();
+            collection.setId(collectionName);
```
ZhijunZhao: These two lines are not used.

yungez: fixed

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
```diff
@@ -202,76 +273,6 @@ public void dropCollection(String collectionName) {
...
-
-        for (int i = 0; i < results.size(); i++) {
-            final T entity = dbConverter.convertFromDocument(results.get(i), entityClass);
-            entities.add(entity);
-        }
-
-        return entities;
-    }
-
     public DocumentDbFactory getDocumentDbFactory() {
```
ZhijunZhao: getDocumentDbFactory and getDocumentDbConverter are not referenced.

yungez: no, they're not referenced.

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
```diff
@@ -158,32 +234,27 @@ public DocumentCollection createCollection(String collectionName, RequestOptions
 
     }
 
-    private DocumentCollection getCollection(String dbName, String collectionName) {
+    private DocumentCollection createOrGetCollection(String dbName, String collectionName) {
```
ZhijunZhao: A few functions like this don't have the logging code.

ZhijunZhao:
It might be good to add a new API to DocumentDbRepository, as suggested by service team. 

FindAllByProperty

spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
yungez: nice. add missed logger info to write operations

yungez:
there're lots of feature could be supported next, e.g paging, findall, delete, customized find, count, batch operation together with findByProperty. timeline is not determined yet based on roadmap.

PR #28, opened 2017-07-12
Spring Boot Starter for Service Bus

ZhijunZhao:
@yungez @aovechki


azure-servicebus-spring-boot-autoconfigure/src/main/java/com/microsoft/azure/autoconfigure/servicebus/ServiceBusAutoConfiguration.java
```diff
@@ -32,13 +34,18 @@ public ServiceBusAutoConfiguration(ServiceBusProperties properties) {
...
+        if (properties.getConnectionString() != null && properties.getQueueName() != null &&
+                properties.getQueueReceiveMode() != null) {
+            return new QueueClient(new ConnectionStringBuilder(properties.getConnectionString(),
+                    properties.getQueueName()), properties.getQueueReceiveMode());
         }
 
-        if (properties.getQueueConnectionString() == null) {
-            LOG.error("Property azure.servicebus.queue-connection-string is not set.");
+        if (properties.getConnectionString() == null) {
+            LOG.error("Property azure.servicebus.connection-string is not set.");
```
aovechki: Looks like the property names are defined as const elsewhere. It'd be better to factor out writing this particular error condition as it happens in many places.

azure-servicebus-spring-boot-starter/README.md
```diff
@@ -0,0 +1,63 @@
...
+For subscription, please specify your created subscription name and receive mode.
+
+```
+azure.servicebus.subscription-name=put-your-subscription-name-here
+azure.servicebus.subscription-receive-mode=peeklock
+```
+
+### Add auto-wiring code
+
+Add below alike code to autowire the clients you want to use in your application. You can reference the sample code in azure-servicebus-spring-boot-starter-sample folder.
```
aovechki: Consider using the following text:

You can use the following code to autowire the Azure Service Bus Queue, Topic, and Subscription clients in your Spring Boot application. Please see sample code in the [azure-servicebus-spring-boot-starter-sample](../azure-servicebus-spring-boot-starter-sample) folder as a reference.

azure-servicebus-spring-boot-autoconfigure/src/main/java/com/microsoft/azure/autoconfigure/servicebus/ServiceBusAutoConfiguration.java
ZhijunZhao: Do you think it is a good idea to add pubic data member like below to ServiceBusProperties?

public static final String CONNECTION_STRING_PROPERTY = "azure.servicebus.connection-string";

PR #30, opened 2017-07-13
Gitignore

rajadilipkolli:
Adds gitignore at project level to ignore unnecessary files while checking


ZhijunZhao:
Hi Raja, how are these .checkstyle files generated? What's the using scenario? We've used the maven-checkstyle-plugin plugin.

Besides, if we plan to ignore this file, I'll just ignore it in the root .gitignore.

rajadilipkolli:
All these checkstyles are generated when I import the project as maven project in eclipse oxygen when I installed the plugin it it has poped up.

I agree that we should keep at global level but I still got generated and showing when I try to commit.

ZhijunZhao:
I can reproduce your issue now. Please update your PR with below change and it works on my side. 

capture

rajadilipkolli:
closing in favour of #34

PR #31, opened 2017-07-17
set userAgentSuffix for documentDb starters

yungez:
@ZhijunZhao  @aovechki


azure-documentdb-spring-boot-autoconfigure/src/main/java/com/microsoft/azure/autoconfigure/documentdb/DocumentDBAutoConfiguration.java
```diff
@@ -30,6 +30,7 @@
 @EnableConfigurationProperties(DocumentDBProperties.class)
 public class DocumentDBAutoConfiguration {
     private static final Logger LOG = LoggerFactory.getLogger(DocumentDBAutoConfiguration.class);
+    private static final String UserAgentSuffix = "azure-documentdb-spring-boot-starter/0.1.3";
```
ZhijunZhao: Maybe we can use the jar file name and get it during run time.

yungez: constant name check not applying to private field. still update it to upper case for better looking

PR #34, opened 2017-07-17
Spring Boot Starter For Azure Media Services

rajadilipkolli:
adds Spring Boot Starter For Azure Media Services
adds checkstyle ,factorypath, springBeans ,bin to git ingore to prevent unnecessary files being commited 

Using this instead of #30 

cc @yungez @xscript @aovechki


azure-media-spring-boot-autoconfigure/src/main/java/com/microsoft/azure/autoconfigure/azuremedia/MediaServicesProperties.java
```diff
@@ -0,0 +1,96 @@
...
+    private String mediaServiceUri;
+    @NotNull
+    private String oAuthUri;
+    @NotNull
+    private String clientId;
+    @NotNull
+    private String clientSecret;
+    @NotNull
+    private String scope;
+
```
ZhijunZhao: Although the SDK expects client id and client secret, I prefer the name `accountName` and `accountKey`, for that're the names shown in Azure portal.

1. Please only apply not null validation to the accountName and accountKey.
2. For the rest three variables, assign the default value if null.

azure-media-spring-boot-starter-sample/src/main/resources/application.properties
```diff
@@ -0,0 +1,5 @@
+azure.media.mediaServiceUri=https://media.windows.net/API/
+azure.media.oAuthUri=https://wamsprodglobal001acs.accesscontrol.windows.net/v2/OAuth2-13
+azure.media.clientId=put-your-azure-media-clientId-here
+azure.media.clientSecret=put-your-azure-media-secret-here
```
ZhijunZhao: let's use the naming convention as below, to be consistent with other starters. 

azure.media.account-name=
azure.media.account-key=

Remove the rest three properties from the property file.

azure-media-spring-boot-starter/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Open `application.properties` file and add below property with your Azure Storage connection string.
+
+```
+azure.media.mediaServiceUri=https://media.windows.net/API/
+azure.media.oAuthUri=https://wamsprodglobal001acs.accesscontrol.windows.net/v2/OAuth2-13
+azure.media.clientId=put-your-azure-media-clientId-here
+azure.media.clientSecret=put-your-azure-media-secret-here
+azure.media.scope=urn:WindowsAzureMediaServices
+```
```
ZhijunZhao: update here accordingly.

ZhijunZhao:
@rajadilipkolli really appreciate your work on this.

rajadilipkolli:
@ZhijunZhao  I will incorporate the changes and update the PR. I will do it when I get free cycles it might take couple of days

azure-media-spring-boot-autoconfigure/src/test/java/com/microsoft/azure/autoconfigure/azuremedia/MediaServicesPropertiesTest.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final String MEDIA_SERVICE_PROPERTY = "azure.media.mediaServiceUri";
+    private static final String OAUTH_URI = "some oauth URI";
+    private static final String OAUTH_URI_PROPERTY = "azure.media.oAuthUri";
+    private static final String CLIENTID = "some clientId";
+    private static final String CLIENTID_PROPERTY = "azure.media.clientId";
+    private static final String CLIENTSECRET = "some clientSecret";
+    private static final String CLIENTSECRET_PROPERTY = "azure.media.clientSecret";
+    private static final String SCOPE = "some oauth URI";
+    private static final String SCOPE_PROPERTY = "azure.media.scope";
+
```
rajadilipkolli: Can you please elaborate more on what you are looking for

azure-media-spring-boot-starter-sample/src/main/resources/application.properties
rajadilipkolli: To be inline with other changes how about using?

azure.mediaservices.account-name=
azure.mediaservices.account-key=

azure-mediaservices-spring-boot-autoconfigure/src/main/java/com/microsoft/azure/autoconfigure/mediaservices/MediaServicesAutoConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+    private MediaContract createMediaContract() {
+        LOG.debug("createMediaContract called");
+        final com.microsoft.windowsazure.Configuration configuration = MediaConfiguration
+                .configureWithOAuthAuthentication(
+                        mediaServicesProperties.getMediaServiceUri(),
+                        mediaServicesProperties.getoAuthUri(),
+                        mediaServicesProperties.getAccountName(),
+                        mediaServicesProperties.getAccountKey(),
+                        mediaServicesProperties.getScope());
+        return MediaService.create(configuration);
```
rajadilipkolli: azure media services API is allowing to create mediacontract with any dummy value, it only validates when original service is invoked.

Ideally if azure mediaservices spring boot starter creates a bean that should be correct. Can you help me how to solve this?

rajadilipkolli: @ZhijunZhao appreciate your help

rajadilipkolli: I am thinking of adding a call to service as below, this makes sure that everything is valid. your suggestions are welcome
```
 final MediaContract mediaService = MediaService.create(configuration);
        mediaService.create(AccessPolicy.create("noAccessPolicy", 0, EnumSet.of(AccessPolicyPermission.NONE)));
        return mediaService;
```

ZhijunZhao: Exceptions should be thrown if user specify invalid property setting. The question is whether it's better to throw the exception in our package or do it in user's code. Right now, we couldn't provide a better way to verify and tell user that their account name / key is invalid. 

So, for simplicity, let's leave it as it is.

PR #35, opened 2017-07-17
Use TestConfiguration instead of Configuration

rajadilipkolli:
TestConfiguration is specially designed for JUNIT testing , hence using instead of @Configuration


ZhijunZhao:
@rajadilipkolli thanks for you PR. Below is what I find for the difference between these two annotations. 

> If you want to customize the primary configuration, you can use a nested
`@TestConfiguration` class. Unlike a nested `@Configuration` class which would be used
instead of your application's primary configuration, a nested `@TestConfiguration` class
will be used in addition to your application's primary configuration.

In our test cases, the configuration of that Config class is all we need. My understanding is that `@Configuration` meets exactly what we need.

Besides, if you search the source code of spring boot, you can only find the usage of `@Configuration` for inner config class in their tests. 

https://github.com/spring-projects/spring-boot

rajadilipkolli:
@ZhijunZhao : Agreed, thank you for correcting me

PR #39, opened 2017-07-19
Support Network Proxy for Azure Media Services

rajadilipkolli:
Updates ReadMe to add azure media services starter
Polish code

### DIFF
--- a/azure-mediaservices-spring-boot-starter/README.md
+++ b/azure-mediaservices-spring-boot-starter/README.md
@@ -23,7 +23,7 @@ azure.mediaservices.account-key=put-your-media-services-account-key-here
 ```
 
 # Optional
-If you are using network proxy then add below properties to run azure media services behind proxy
+If you are using network proxy then add below properties to run azure media services behind proxy.
 
 ```
 azure.mediaservices.proxy-host=put-your-network-proxy-host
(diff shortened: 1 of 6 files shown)


azure-mediaservices-spring-boot-starter/README.md
```diff
@@ -22,6 +22,15 @@ azure.mediaservices.account-name=put-your-media-services-account-name-here
 azure.mediaservices.account-key=put-your-media-services-account-key-here
 ```
 
+# Optional
+If you are using network proxy then add below properties to run azure media services behind proxy
```
ZhijunZhao: Missing the period at the end of sentence. 

BTW, are you a user of the Media Services? If yes, what's the reason for using proxy in your case?

rajadilipkolli: Yeah I am user of Media Services, my client has a strict Firewall policy . we cannot communicate to Azure Media Services directly and it can be done only using network proxy provided by client

rajadilipkolli:
@ZhijunZhao : if supported shouldn't we extend network proxy to all other starters as optional feature 

My intension is no one should face issues like I have faced , I found solution after banging my head 15 days 
https://stackoverflow.com/questions/39412377/how-to-use-network-proxy-when-connecting-to-microsoft-azure-media-services

PR #52, opened 2017-07-25
add travis, codecov, license info

yungez:
@ZhijunZhao, @xscript,  @aovechki

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,12 @@
 language: java
 jdk:
   - oraclejdk8
-script: mvn clean verify
+script: "mvn cobertura:cobertura"
 sudo: required
 cache:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,8 +1,12 @@
 language: java
 jdk:
   - oraclejdk8
-script: mvn clean verify
+script: "mvn cobertura:cobertura"
```
ZhijunZhao: Does this command include the `mvn verify` phase? I think findbugs is run during the mvn verify phase.

xscript: @ZhijunZhao Before this command, Travis will run `mvn install` automatically. It will cover the `verify` phase.

PR #55, opened 2017-07-25
Use parent project version for starters dynamically

rajadilipkolli:
This will be helpful more while releasing the version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
         </developer>
     </developers>
 	
-	<organization>
+    <organization>
     	<name>Microsoft</name>
     	<url>https://www.microsoft.com</url>
     </organization>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -33,7 +33,11 @@
             <email>yungez@microsoft.com</email>
         </developer>
     </developers>
-
+	
+	<organization>
+    	<name>Microsoft</name>
+    	<url>https://www.microsoft.com</url>
+    </organization>
```
ZhijunZhao: Could you fix the indentation to make the format better?

rajadilipkolli: Fixed

ZhijunZhao:
@rajadilipkolli Thanks for this change!

PR #61, opened 2017-07-31
Yungez integrationtest

yungez:
add integration test for spring data document db. 
@xscript  @aovechki @ZhijunZhao

### DIFF
--- a/azure-spring-boot-starter-parent/pom.xml
+++ b/azure-spring-boot-starter-parent/pom.xml
@@ -137,7 +137,7 @@
                 </dependencies>
                 <executions>
                     <execution>
-                        <phase>verify</phase>
+                        <phase>compile</phase>
                         <goals>
                             <goal>check</goal>
                         </goals>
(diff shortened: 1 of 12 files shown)


azure-spring-boot-starter-parent/pom.xml
```diff
@@ -137,7 +137,7 @@
                 </dependencies>
                 <executions>
                     <execution>
-                        <phase>verify</phase>
+                        <phase>compile</phase>
```
yungez: findbugs and cobertura-integration-test has conflict because of Cobertura 2.7 alters project.build.outputDirectory in order to produce report for integration tests (new kind of report introduced in 2.7). In the next step Findbugs checks violations not on just compiled classes but on instrumented. https://stackoverflow.com/questions/29594445/cobertura-maven-plugin-conflicts-with-findbugs
this change is to run findbug at compile phase instead of verify to workaround the issue.

yungez:
Leave a comment

PR #74, opened 2017-08-01
adds maven wrapper

rajadilipkolli:
Adding maven wrapper from https://github.com/takari/maven-wrapper eradicates users to setup maven in local environment and can start developing or contributing directly

@xscript @yungez @aovechki @ZhijunZhao

PR #75, opened 2017-08-01
removes unnecessary sudo

rajadilipkolli:
removes sudo from travis build which is not necessary

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ jdk:
 
 script: "mvn cobertura:cobertura"
 
-sudo: required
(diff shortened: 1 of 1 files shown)


rajadilipkolli:
Incase if #74  is not accepted then this is a must

rajadilipkolli:
@yungez can I assume that #74 will not be accepted?

yungez:
@rajadilipkolli  thanks for help add maven wrapper,  I think i missed to submit comments on #74, once the comments fixed, will approve and merge it PR #74 .

PR #95, opened 2017-08-03
Azure AD integration starter

yaweiw:
* filter, autoconfigure and properties for Azure AD integration starter
* units


yaweiw:
@xscript @yungez @ZhijunZhao @zhzy0077 @aovechki

PR #96, opened 2017-08-03
defaulting to singleton Bean scope

rajadilipkolli:
As only one client connection should be established using prototype is not correct , hence reverting to default singleton scope.
cc @xscript @yungez @aovechki @ZhijunZhao


ZhijunZhao:
@rajadilipkolli any issue you found using prototype?

rajadilipkolli:
Its establishing new client connection every time Mediaclient is called which is heavy weight and time consuming also from memory perspective it is increasing drastically

I think same applies to storage autoconfigure as well where it is defined as prototype

`Singleton, will create a new instance in the first call, and return it in subsequent calls. Prototype will return a new instance each time. ... If scope is set to singleton, the Spring IoC container creates exactly one instance of the object defined by that bean definition` So as one instance is enough using the same

rajadilipkolli:
I feel https://github.com/Microsoft/azure-spring-boot-starters/blob/master/storage/azure-storage-spring-boot-autoconfigure/src/main/java/com/microsoft/azure/autoconfigure/storage/StorageAutoConfiguration.java#L39 this should also be changed to singleton. I don't see reason for prototype

ZhijunZhao:
@rajadilipkolli The client is not guaranteed to be thread-safe. If user want to use multi threading and multiple client instances, it has to be prototype.

Issue #53, opened 2017-07-25
errors from maven-javadoc-plugin plugin 


ZhijunZhao:
[ERROR] MavenInvocationException: Error when invoking Maven, consult the invoker log file: D:\azure-spring-boot-starters\azure-mediaservices-spring-boot-starter-sample\target\invoker\maven-javadoc-plugin1034693639.txt
[WARNING] Creating fake javadoc directory to prevent repeated invocations: D:\azure-spring-boot-starters\azure-mediaservices-spring-boot-starter\target\apidocs
[ERROR] Error fetching link: D:\azure-spring-boot-starters\azure-mediaservices-spring-boot-starter\target\apidocs/package-list. Ignored it.

PR #101, opened 2017-08-07
use BOM module as the parent of sample modules

### LINKED ISSUE #93: using azure-spring-boot-starter-parent as parent in standalone sample project met error
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (validate) on project demo-documentdb: Failed during checkstyle execution: Unable to find configuration file at location: ../config/checkstyle.xml: Could not find resource '../config/checkstyle.xml'. -> [Help 1]

version 0.1.4

ZhijunZhao:
@yungez @xscript @aovechki 

fix #93 #53


common/azure-spring-boot-starter-parent/pom.xml
```diff
@@ -225,8 +225,6 @@
                     <instrumentation>
                         <excludes>
                             <exclude>com/microsoft/azure/spring/common/GetHashMac.class</exclude>
```
yungez: one concern on parent-pom is that if user used parent package as parent, build will break.

Issue #87, opened 2017-08-02
SQL injection vulnerability


zhzy0077:
Line 191 of `DocumentDbTemplate` class
```java
final String query = "SELECT * FROM root r WHERE r.id='" + dbName + "'";```

Is it vulnerable of SQL injection?

PR #110, opened 2017-08-10
enable appveyor

yungez:
1. enable appveyor, documentdb integration test against emulator
2. fix issue #87 

@yungez @xscript @aovechki @zhzy0077


documentdb/spring-data-azure-documentdb/src/test/java/com/microsoft/azure/spring/data/documentdb/repository/core/DocumentDbTemplateIT.java
```diff
@@ -16,25 +16,31 @@
...
+@RunWith(SpringJUnit4ClassRunner.class)
 @PropertySource(value = {"classpath:application.properties"})
 public class DocumentDbTemplateIT {
 
-    private static final String DOCUMENTDB_URL_PROPERTY_NAME = "documentdb.uri";
-    private static final String DOCUMENTDB_KEY_PROPERTY_NAME = "documentdb.key";
     private static final String TEST_DB_NAME = "testdb";
     private static final Person TEST_PERSON = new Person("testid", "testfirstname", "testlastname");
+    
+    @Value("${documentdb.uri}")
```
ZhijunZhao: Can you try @Value("${DOCUMENTDB_URI}")? I think it works and the application.properties file won't be needed.

yungez: application.properties is used in 2 scenarios:
1. when testing on azure, after azure document db created during test exection, ant-plugin will write db uri and key into application.properties, then test will use it. ant-plug is not able to set environment variable
2. when testing on emulator, in this case, emulator uri is set before test execution.

PR #115, opened 2017-08-14
add more integration and unit test for documentdb

yungez:
@ZhijunZhao @aovechki  @yaweiw


documentdb/spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
```diff
@@ -99,7 +100,7 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
                 return null;
             }
         } catch (DocumentClientException e) {
-            throw new RuntimeException(e.getMessage());
+            throw new RuntimeException("findById exception: " + e.getMessage());
```
ZhijunZhao: Use this constructor?

RuntimeException(String message, Throwable cause)
Constructs a new runtime exception with the specified detail message and cause.

yungez: resolved

PR #119, opened 2017-08-16
add deletebyId, more tests in documentdb

yungez:
add deletebyId, more tests in documentdb.

@ZhijunZhao @aovechki @yaweiw

### DIFF
--- a/common/azure-spring-boot-starter-parent/pom.xml
+++ b/common/azure-spring-boot-starter-parent/pom.xml
@@ -127,6 +127,9 @@
                 <configuration>
                     <show>private</show>
                     <failOnError>false</failOnError>
+                    <sourceFileExcludes>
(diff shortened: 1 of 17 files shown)


common/azure-spring-boot-starter-parent/pom.xml
```diff
@@ -127,6 +127,9 @@
                 <configuration>
                     <show>private</show>
                     <failOnError>false</failOnError>
+                    <sourceFileExcludes>
+                        <exclude>com/microsoft/azure/spring/data/documentdb/core/mapping/BasicDocumentDbPersistentProperty.java</exclude>
+                    </sourceFileExcludes>
```
yungez: exclude this to avoid warning. if public property with name like is*Property, javadoc will warn.
https://bugs.launchpad.net/ubuntu/+source/openjdk-8/+bug/1706969
https://bugs.openjdk.java.net/browse/JDK-8185424

PR #121, opened 2017-08-16
Spring Social support for Microsoft Graph

ZhijunZhao:
@yungez @xscript @aovechki 

The starter provides very limited API binding, email sending and receiving only, for Microsoft Graph API. But it provides the capability that allows users to customize their own API bindings.


microsoft-graph/microsoft-graph-spring-social-starter-sample/src/main/java/sample/microsoft/graph/HelloController.java
```diff
@@ -36,4 +41,17 @@ public String helloFacebook(Model model) {
...
+        if (connectionRepository.findPrimaryConnection(Microsoft.class) == null) {
+            return "redirect:/connect/microsoft";
+        }
+
+        final RestTemplate restTemplate = microsoft.customOperations().getRestTemplate();
+        final URI uri = microsoft.customOperations().getGraphAPIURI("me/contacts");
+        final Contacts contacts = restTemplate.getForObject(uri, Contacts.class);
+        model.addAttribute("contacts", contacts.getContacts());
+
+        return "contacts";
```
yungez: return this string literally?

microsoft-graph/microsoft-graph-spring-social-starter-sample/src/main/resources/application.properties
```diff
@@ -1,2 +1,3 @@
 spring.social.microsoft.app-id=put-your-app-id-here
 spring.social.microsoft.app-secret=put-your-app-secret-here
+logging.level.sample:debug
```
yungez: remove this?

microsoft-graph/microsoft-graph-spring-social-starter-sample/src/main/java/sample/microsoft/graph/HelloController.java
ZhijunZhao: yes, it would be mapped to contacts.html

PR #126, opened 2017-08-25
Polish

rajadilipkolli:
adds generics
removes deprecated package
removes redundant groupId and version  from maven which is managed by parent POM
uses Static Instead of reference
removes unused logger


activedirectory/azure-ad-integration-spring-boot-autoconfigure/src/main/java/com/microsoft/azure/autoconfigure/aad/UserPrincipal.java
```diff
@@ -155,21 +157,21 @@ public void verify(JWTClaimsSet claimsSet, SecurityContext ctx) throws BadJWTExc
 
     private List<UserGroup> loadUserGroups(String graphApiToken) throws Exception {
         final String responseInJson = AzureADGraphClient.getUserMembershipsV1(graphApiToken);
-        final List<UserGroup> userGroups = new ArrayList<UserGroup>();
+        final List<UserGroup> lUserGroups = new ArrayList<>();
```
ZhijunZhao: could you keep the previous coding style?

if changed to `lUserGroups`, a lot of other variable names have to be changed.

cloudfoundry/azure-cloud-foundry-service/src/main/java/com/microsoft/azure/cloudfoundry/service/VcapResult.java
```diff
@@ -205,9 +206,9 @@ private void addOrReplace(MutablePropertySources propertySources,
                     .get(PROPERTY_SOURCE_NAME);
             if (source instanceof MapPropertySource) {
                 target = (MapPropertySource) source;
-                for (final String key : map.keySet()) {
-                    if (!target.containsProperty(key)) {
-                        target.getSource().put(key, map.get(key));
+                for (final Entry<String, Object> entry : map.entrySet()) {
+                    if (!target.containsProperty(entry.getKey())) {
+                        target.getSource().put(entry.getKey(), map.get(entry.getKey()));
```
ZhijunZhao: Could you elaborate the issue here?

rajadilipkolli: From performance perspective it is always better to use entryset() for iterating HashMap than keyset().

For more information on this refer [stackoverflow] (https://stackoverflow.com/a/3870093/5557885)

activedirectory/azure-ad-integration-spring-boot-autoconfigure/src/main/java/com/microsoft/azure/autoconfigure/aad/UserPrincipal.java
rajadilipkolli: There is a class level variable userGroups and the same name is being used locally inside this method which is causing readability issue and to distinguish both variables I have changed this. If i revert back there can u suggest name for class level variable?

PR #130, opened 2017-08-28
Polish active directory auto-configuration

snicoll:
This commit makes sure that the HTTP filter is not created in a non-web based Spring Boot application.


activedirectory/azure-ad-integration-spring-boot-autoconfigure-sample/pom.xml
```diff
@@ -171,17 +155,6 @@
...
-                <groupId>org.springframework.boot</groupId>
-                <artifactId>spring-boot-maven-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <goals>
-                            <goal>repackage</goal>
-                        </goals>
-                    </execution>
-                </executions>
-            </plugin>
```
ZhijunZhao: With this plugin, we can run the app with below command. Why removing it here?

java -jar target\azure-ad-integration-spring-boot-autoconfigure-sample-0.0.1-SNAPSHOT.jar

activedirectory/azure-ad-integration-spring-boot-autoconfigure-sample/pom.xml
```diff
@@ -113,14 +105,6 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <configuration>
-                    <source>${java.version}</source>
-                    <target>${java.version}</target>
-                </configuration>
-            </plugin>
```
ZhijunZhao: why removing this plugin?

activedirectory/azure-ad-integration-spring-boot-autoconfigure-sample/pom.xml
ZhijunZhao: @snicoll BTW, could you sign the CLA? :)

activedirectory/azure-ad-integration-spring-boot-autoconfigure-sample/pom.xml
snicoll: It is already provided by the parent. This is 100% copy/paste

activedirectory/azure-ad-integration-spring-boot-autoconfigure-sample/pom.xml
snicoll: Idem. Remove it and you'll see it's still generated.

ZhijunZhao: Yes, it's generated, but it would have below error.

no main manifest attribute, in target\azure-ad-integration-spring-boot-autoconfigure-sample-0.0.1-SNAPSHOT.jar

snicoll: uh?

activedirectory/azure-ad-integration-spring-boot-autoconfigure-sample/pom.xml
ZhijunZhao: For all the sample projects, their parents are actually the BOM module. So, it's not provided.

snicoll: That's not what Git says: https://github.com/Microsoft/azure-spring-boot-starters/pull/130/files#diff-ade6d2380d8cccbe2e84a5f782eea4a0R8

activedirectory/azure-ad-integration-spring-boot-autoconfigure-sample/pom.xml
ZhijunZhao: https://docs.spring.io/spring-boot/docs/current/maven-plugin/repackage-mojo.html

activedirectory/azure-ad-integration-spring-boot-autoconfigure-sample/pom.xml
ZhijunZhao: My bad. You're right. 👍

ZhijunZhao: Now I remember why I added this plugin specifically. 

The reason is that the default java version specified by starter-parent is 1.6. You can check the pom definition here. 

https://search.maven.org/remotecontent?filepath=org/springframework/boot/spring-boot-starter-parent/1.5.6.RELEASE/spring-boot-starter-parent-1.5.6.RELEASE.pom

In our sample code, we used some Java 8 feature. So we need to set it to 1.8. Or else compilation would fail like below.

https://travis-ci.org/Microsoft/azure-spring-boot-starters/builds/269546952

ZhijunZhao: @snicoll One think I spent more than one hour to figure out but failed is that where this plugin `maven-compiler-plugin` is added. 

I choose the `show effect POM` option. Yes, this plugin is actually shown. But I searched `spring-boot-starter-parent` and its parent `spring-boot-dependencies`, but couldn't find this plugin. The only thing I find is its dependency management in `spring-boot-dependencies`.

Stéphane, could you help to explain how this plugin is magically added to the final pom?

snicoll: Just set `java.version` to 1.8. There is no need to copy/paste the plugin configuration for that.

snicoll: It doesn't have to be set as we override the standard properties of the plugin with `java.version`: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-starters/spring-boot-starter-parent/pom.xml#L24

ZhijunZhao:
@snicoll  Any update? I can go ahead and update your branch if you don't mind.

snicoll:
Sure go ahead.

PR #138, opened 2017-08-29
Polish BOM module

### LINKED ISSUE #128: BOM should not provide dependency management for third party libraries
It looks like you attempted to do something similar to `spring-boot-dependencies` which is the BOM of a "platform" (with stronger semantics). This project is more narrowed and its bom should only expose dependency management for the artifacts that the project produces (so in this case, azure artifacts). It should not have an opinion about third party dependencies (otherwise including that BOM in a project potentially overrides decisions that were made about them)

For versions that match the version of the project (everything in an ideal world except maybe Microsoft-specific jars that the starters require), I'd set them explicitly rather than using `${bom.version}`. If you're including the bom for Azure X.Y.Z, it should be X.Y.Z. and nothing else (Maven does not allow to override that property, which is good so it's not strictly necessary but the intent is more clear if you hardcode the current version).

Here is an example of BOM for reference: http://repo.spring.io/release/org/springframework/spring-framework-bom/4.3.8.RELEASE/spring-framework-bom-4.3.8.RELEASE.pom

ZhijunZhao:
fix #128 

@yungez @xscript @aovechki

### DIFF
--- a/activedirectory/azure-ad-integration-spring-boot-autoconfigure/pom.xml
+++ b/activedirectory/azure-ad-integration-spring-boot-autoconfigure/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <groupId>com.microsoft.azure</groupId>
         <artifactId>azure-spring-boot-starter-parent</artifactId>
-        <version>0.1.6-beta</version>
+        <version>0.1.6-SNAPSHOT</version>
         <relativePath>../../common/azure-spring-boot-starter-parent/pom.xml</relativePath>
     </parent>
 
--- a/activedirectory/azure-ad-integration-spring-boot-autoconfigure/pom.xml
+++ b/activedirectory/azure-ad-integration-spring-boot-autoconfigure/pom.xml
@@ -42,11 +42,6 @@
 
     <properties>
         <azure.adal4j.version>1.2.0</azure.adal4j.version>
-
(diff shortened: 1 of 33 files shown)


activedirectory/azure-ad-integration-spring-boot-autoconfigure/pom.xml
```diff
@@ -40,6 +40,10 @@
         <url>https://github.com/Microsoft/azure-spring-boot-starters/tree/master</url>
     </scm>
 
+    <properties>
+        <azure.adal4j.version>1.2.0</azure.adal4j.version>
+    </properties>
```
ZhijunZhao: make the version of Azure package a property so that it can be overwritten

rajadilipkolli: @ZhijunZhao ordering dependencies alphabetically can improve readability and also adding new dependencies later. if you agree I will propose a PR

ZhijunZhao: That would be fantastic! But it would be better after this PR is merged, which should be early tomorrow morning.

rajadilipkolli: I will submit in next couple of days once these spring boot initializr things completes.

PR #140, opened 2017-08-30
merge auto-configure modules with starters

### LINKED ISSUE #131: Missing activedirectory starter?
There is no starter for the active directory integration. That's fine if you don't but currently the `autoconfigure` module brings a bunch of dependencies so it isn't an `autoconfigure` module anymore. 

1. Either that's the dependency that one is supposed to add when they want to get started with active directory and in that case I'd recommend to rename that to starter to clarify the intend
2. Either another starter or that dependency + required dependencies should be added. I'd argue that ordering dependencies by scope makes it clearer. Right now this module brings automatically `spring-web`, `spring-security`, `jackson`, `powermock` (I guess it's missing a `test` scope!). The sample brings more dependencies and it isn't clear what's optional and what's required IMO

ZhijunZhao:
Fix #131

@yungez @xscript @aovechki


ZhijunZhao:
Fix #131

Issue #160, opened 2017-09-06
Keyvault has no configuration metadata


snicoll:
The keyvault Spring Boot integration defines a number of properties but they have no metadata so no IDE inspection. I see why you don't have a `@ConfigurationProperties` object for this feature (since adding the `KeyVaultPropertySource` to the `Environment` is happening very early) but you can write manual meta-data for those.

Issue #161, opened 2017-09-06
azure.token.acquire.timeout.seconds is wrongly namespaced


snicoll:
If this property is keyvault specific, it should be under the `azure.keyvault` namespace IMO. If it isn't, I'd still rename it as the intend of the prefix is to define an origin for the feature rather than translating the property name itself.

yungez:
thanks @snicoll Stephane for explaning that prefix is to define the origin of feature. No, those properties are not keyvault specific. Will rename them to follow the naming convention to indicate origin of the feature

yungez:
closed by PR #167

PR #163, opened 2017-09-07
Add keyvault metadata

eddumelendez:
See gh-160


eddumelendez:
I have not included `azure.token.acquire.timeout.seconds` due to not sure if should be renamed. See #161

Issue #144, opened 2017-08-31
DocumentDB: Select query failed when cross partition


yuwzho:
When I use `DocumentDbRepository.findAll()` method to select all entities, exception occurs:
`{"message":"Cross partition query is required but disabled. Please set x-ms-documentdb-query-enablecrosspartition to true, specify x-ms-documentdb-partitionkey, or revise your query to avoid this exception."}`

As [docs](https://docs.microsoft.com/en-us/azure/cosmos-db/partition-data) says, the CosmosDB saves data and index them with partition keys. And when running query, the REST API requires specific the partition key or enable crosspartitionkey flag. But in the [lib](https://github.com/Microsoft/azure-spring-boot-starters/blob/master/documentdb/spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java#L209), it set the options to `null`, make the call failed.

In my thought, there should be two methods:

1. `findAll()` to enable `setEnableCrossPartitionQuery(true)`
2. Another method to select entities with specific partition key.

yuwzho:
If the collection doesn't set a partition key, the CRD operation can work properly, But once the partition key specified, all the CRD request required partition key.

PR #169, opened 2017-09-14
add partition support in docdb

yungez:
@ZhijunZhao @aovechki @yuwzho 

#144

### DIFF
--- a/documentdb/spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
+++ b/documentdb/spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
@@ -54,16 +54,19 @@ public class DocumentDbTemplate implements DocumentDbOperations, ApplicationCont
         this(new DocumentDbFactory(client), mappingDocumentDbConverter, dbName);
     }
 
-
     public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
     }
 
-    public <T> T insert(T objectToSave) {
-        return insert(getCollectionName(objectToSave.getClass()), objectToSave);
+    public <T> T insert(T objectToSave, String partitionKeyFieldValue) {
+        return insert(getCollectionName(objectToSave.getClass()),
+                objectToSave,
+                partitionKeyFieldValue);
     }
 
 
-    public <T> T insert(String collectionName, T objectToSave) {
+    public <T> T insert(String collectionName,
+                        T objectToSave,
+                        String partitionKeyFieldValue) {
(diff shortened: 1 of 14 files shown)


documentdb/spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
```diff
@@ -311,4 +353,24 @@ private String getCollectionLink(String databaseName, String collectionName) {
     private String getDocumentLink(String databaseName, String collectionName, String documentId) {
         return getCollectionLink(databaseName, collectionName) + "/docs/" + documentId;
     }
+
+    private String getPartitionKeyPath(String partitionKey) {
+        return "/" + partitionKey;
+    }
+
+    private RequestOptions getRequestOptions(String partitionKeyValue, Integer requestUnit) {
```
yuwzho: I see all the caller of this function one of partitionKeyValue and requestUnit is null? That means this function always return null even for the partition case?

yungez: when collection is not partitioned, partitionkey=null, this method will return null reuqestoption
when collection is partitioned, partitionkey has value, this method will return requestoption object.
callers of this method is in DocumentDBTemplate.java, methods in template just pass in partitionkey from repository level. if user choose partition, pass partitionkey in, then requestoption will set partitionkey. null is requestUnit, which only need specify when create collection. when operate with document, no need this param, so null. get this logic from documentdb sdk test

documentdb/spring-data-azure-documentdb/src/main/java/com/microsoft/azure/spring/data/documentdb/core/DocumentDbTemplate.java
```diff
@@ -186,9 +181,17 @@ public void deleteAll(String collectionName) {
...
         }
 
+        SqlQuerySpec sqlQuerySpec = new SqlQuerySpec("SELECT * FROM root c");
+        if (partitionKeyFieldName != null && !partitionKeyFieldName.isEmpty()) {
+            sqlQuerySpec = new SqlQuerySpec("SELECT * FROM root c WHERE c." + partitionKeyFieldName + "=@partition",
+                    new SqlParameterCollection(new SqlParameter("@partition", partitionKeyFieldValue)));
+        }
+
+        final FeedOptions feedOptions = new FeedOptions();
+        feedOptions.setEnableCrossPartitionQuery(true);
```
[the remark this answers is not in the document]
yungez: yes, always true no matter with or without partition

PR #173, opened 2017-09-18
Add Microsoft librarires to bom

### LINKED ISSUE #171: Add Microsoft librarires to bom
Relates to #128

The essence of the bom is to provide dependency management for "your own modules". Strictly speaking, the starters and artifacts that are managed here. However, it is clear you have the authority of having an opinion about Azure related jars as well (since that decision doesn't exist elsewhere). 

I think it would help users if artifacts such as `adal4j`, `azure-keyvault`, `azure-storage`, `azure-media`, `azure-servicebus` were added. To some extend I'd assume that dependency management for `spring-social-microsoft-graph` and `spring-data-azure-document-db`.

If this is accepted, the artifactId of the boml has to change as `azure-spring-boot-starters-bom` does not  reflect the goal anymore. Spring should definitely be there as the bom provides dependency management for the Spring Support. Perhaps something like `azure-spring-boot-bom` would be better? 

If we extend that though a bit, this repository could become the place for Spring Boot support for Azure. Maybe `azure-spring-boot` would therefore a better name for this repository?

ZhijunZhao:
Fix #171

@snicoll @yungez @xscript @aovechki @yaweiw


snicoll:
LGTM

ZhijunZhao:
@yungez could you help to update the appveyor setting due to the repo renaming? It seems the appveyor is running under your own account. Is there anyway we can improve it?

yungez:
@ZhijunZhao, pls see #174. currently no group or public appveyor account we have, so it's under my account. I think travis under personal account also

PR #175, opened 2017-09-19
move samples to a shared location

### LINKED ISSUE #159: Consider moving samples to a shared location
Relates to #132

There is some noise in each starter currently:

1. Most if not all `properties` (the two maven one can be replaced by `java.version` btw)
2. `licenses`
3. `developers`
4. `scm`
5. `dependencyManagement`
6.  The `checkstyle`, `findbugs` and `spring-boot` plugin (As for the latter, we already discussed that it's a duplication since you inherit from Spring Boot's parent).

You could have a `azure-samples` root directory with all the samples in them. That way they are easier to find and you can move that noise to the parent.

ZhijunZhao:
fix #159

@snicoll @yungez @xscript @aovechki @yaweiw


azure-samples/azure-documentdb-spring-boot-autoconfigure-sample/pom.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.microsoft.azure</groupId>
+        <artifactId>azure-samples</artifactId>
```
yungez: all samples share same artifactid? it's confusing..

azure-samples/azure-keyvault-secrets-spring-boot-starter-sample/src/main/java/com/microsoft/azure/keyvault/spring/boot/sample/SampleApplication.java
```diff
@@ -29,7 +29,7 @@ public static void main(String[] args) {
 
     public void run(String... varl) throws Exception {
         LOGGER.info("property spring.datasource.url in Azure Key Vault: {}", dbUrl);
-        LOGGER.info("property yourSecretPropertyName in Azure Key Vault: {}", mySecretProperty);
+        LOGGER.info("property yourSecretPropertyName Azure Key Vault: {}", mySecretProperty);
```
yungez: typo?

PR #177, opened 2017-09-19
put all autoconfigure code into one module

yungez:
1. put all autoconfigure code into one module: aad, documentdb, microsoft graph, storage, service bus, media service
2. delete spring data documentdb code, use dependency to spring-data-cosmosdb-documentdb
3. remove documentdb emulator setup in appveyor
4. update dependency of related starters who contains autoconfigure only


yungez:
@ZhijunZhao, all good points. Can we make it in next PR? this one is already too hugh. :)

PR #178, opened 2017-09-19
Fix sample definition

snicoll:
The `repackage` execution is inherited from the parent so there is no
need to duplicate it here.

@ZhijunZhao this related to our discussion in #159. If you generate an empty app on start.spring.io, you'll get the best practices that we are currently recommending.


ZhijunZhao:
@snicoll, thanks! You're making a valid point. I must have missed something back then.

PR #201, opened 2017-09-28
Polish

rajadilipkolli:
use parent version so that change of version is done only at one place.

### DIFF
--- a/azure-spring-boot-bom/pom.xml
+++ b/azure-spring-boot-bom/pom.xml
@@ -61,42 +61,42 @@
             <dependency>
                 <groupId>com.microsoft.azure</groupId>
                 <artifactId>azure-spring-boot-starter</artifactId>
-                <version>0.1.9-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


azure-spring-boot-bom/pom.xml
```diff
@@ -61,42 +61,42 @@
             <dependency>
                 <groupId>com.microsoft.azure</groupId>
                 <artifactId>azure-spring-boot-starter</artifactId>
-                <version>0.1.9-SNAPSHOT</version>
+                <version>${project.version}</version>
```
ZhijunZhao: I guess this won't work. ${project.version} will be evaluated to the version of user's module.

ZhijunZhao: You can take a try.

snicoll: Please don't use `${project.version}` in a BOM that is sending the wrong message. It's much more obvious to have the actual version.

PR #206, opened 2017-10-17
trim keyvault uri property value

ZhijunZhao:
@yungez @yaweiw


azure-spring-boot-parent/pom.xml
```diff
@@ -62,11 +62,6 @@
                 <configuration>
                     <show>private</show>
                     <failOnError>false</failOnError>
-                    <sourceFileExcludes>
-                        <exclude>
-                            com/microsoft/azure/spring/data/documentdb/core/mapping/BasicDocumentDbPersistentProperty.java
```
yungez: there's no warning anymore?

PR #213, opened 2017-10-19
Remove unused dependency

snicoll:
The auto-configuration module should define only optional dependencies
as much as possible so that the starter brings the necessary bits to
enable a particular feature.

This dependency should not be compiled scope. I've removed it to test
where it is actually used and nothing breaks. Please review and either
restore it with optional scope, adding it with compile scope in the
relevant starter.

### DIFF
--- a/azure-spring-boot/pom.xml
+++ b/azure-spring-boot/pom.xml
@@ -40,12 +40,6 @@
             <artifactId>json</artifactId>
         </dependency>
 
-        <!--Keyvault-->
-        <dependency>
-            <groupId>com.microsoft.rest</groupId>
-            <artifactId>client-runtime</artifactId>
-        </dependency>
-
         <!--Azure active directory-->
         <dependency>
             <groupId>org.springframework</groupId>


snicoll:
Then I'd argue that keyvault should back-off then. You can't force keyvault on everyone so the `EnvironmentPostProcessor` should check for the presence of a configuration key or something before running. If you move the offending code in a separate class that should work just fine.

ZhijunZhao:
@snicoll moving the offending code to a separate class while keeping the same logic magically solves the issue. why?

So, I'll make code changes to make keyvault disabled by default and also move the offending code to a separate class. Will get it done by the end of day.

snicoll:
> moving the offending code to a separate class while keeping the same logic magically solves the issue. why?

As long as you don't expose types in method signature that may not be on the classpath, you are fine. If you check if keyvault should be enabled and that's not the case, not instantiating that class should not trigger any early class loading of the keyvault specific classes. 

Of course, if someone puts that property and the keyvault library is not present, it will blow up at runtime rather than backing off. To prevent that, I suggest that you use `ClassUtils.isPresent()` to make sure that at least keyvault is set when the property has been enabled. 

We have a few corner cases like that with Spring Boot when we can't rely on `@ConditionalOnClass`.

Having said that, I haven't looked deep in the code. My point is that the auto-configure module should not expose the keyvault bit or any other dependency to everyone.

PR #215, opened 2017-10-19
enable keyvault conditionally

### LINKED ISSUE #216: Cloud Foundry sample missing a dependency
The cloud foundry sample app fails to start with this error - java.lang.ClassNotFoundException: com.microsoft.rest.credentials.ServiceClientCredentials, need to add this dependency:

		<dependency>
			<artifactId>azure-client-runtime</artifactId>
			<version>1.0.0</version>
		</dependency>

ZhijunZhao:
@snicoll @yungez @xscript @yaweiw 

Fix #216


azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultProcessor.java
```diff
@@ -6,15 +6,13 @@
...
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.env.EnvironmentPostProcessor;
 import org.springframework.core.env.ConfigurableEnvironment;
-import org.springframework.core.env.MutablePropertySources;
-import org.springframework.core.env.StandardEnvironment;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 
-public class KeyVaultPropertyEnvironmentPostProcessor implements EnvironmentPostProcessor {
+public class KeyVaultProcessor implements EnvironmentPostProcessor {
```
snicoll: I'd call that `KeyVaultEnvironmentPostProcessor`

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultProcessor.java
```diff
@@ -39,24 +37,11 @@ public void postProcessEnvironment(ConfigurableEnvironment environment, SpringAp
...
-                            new KeyVaultPropertySource(kvOperation));
-                } else {
-                    sources.addFirst(new KeyVaultPropertySource(kvOperation));
-                }
-
-            } catch (Exception e) {
-                e.printStackTrace();
-            }
+        if (enabled && ClassUtils.isPresent("com.microsoft.azure.keyvault.KeyVaultClient",
+                null)) {
```
snicoll: Use the ClassLoader's Class rather than `null`. I would also wrap that in a separate `private boolean isKeyVaultClientAvailable()`

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultProcessorHelper.java
```diff
@@ -0,0 +1,49 @@
...
+    public KeyVaultProcessorHelper(ConfigurableEnvironment environment,
+                                   String id, String key, String uri, long timeout) {
+        this.environment = environment;
+        this.id = id;
+        this.key = key;
+        this.uri = uri;
+        this.timeout = timeout;
+    }
+
+    public void addKeyVaultPropertySource() {
```
snicoll: You can extract the properties here since you have the `Environment`. That makes sure the logic is in one place.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+        boolean enabled = true;
+        if (environment.getProperty(Constants.AZURE_KEYVAULT_ENABLED) != null) {
+            enabled = Boolean.parseBoolean(environment.getProperty(Constants.AZURE_KEYVAULT_ENABLED));
+        }
+
+        return enabled && isKeyVaultClientAvailable();
+    }
+
+    private boolean isKeyVaultClientAvailable(){
+        return ClassUtils.isPresent("com.microsoft.azure.keyvault.KeyVaultClient", null);
```
ZhijunZhao: @snicoll Changing null to `ClassLoader.getSystemClassLoader()` doesn't work.

ZhijunZhao: I see null is used a lot in spring boot source code for this `ClassUtils.isPresent` call.

If changing to `ClassLoader.getSystemClassLoader()`,

1. It works well if directly run in Intellij.
2. doesn't work if `java -jar *.jar`

snicoll: I didn't refer to the system classloader. I referred to the `Class`'s `ClassLoader`. So `KeyvaultEnvironmentPostProcessor.class.getClassLoader()`

PR #219, opened 2017-10-20
Polish

snicoll:


anujku:
@snicoll Why was the class access modifier for `KeyVaultEnvironmentPostProcessorHelper` removed ?

PR #232, opened 2017-11-17
Spring Boot 2.0 Support

ZhijunZhao:
@yungez @xscript @yaweiw

### DIFF
--- a/azure-spring-boot-samples/azure-active-directory-spring-boot-sample/pom.xml
+++ b/azure-spring-boot-samples/azure-active-directory-spring-boot-sample/pom.xml
@@ -33,6 +33,7 @@
         <dependency>
             <groupId>org.springframework.security.oauth</groupId>
             <artifactId>spring-security-oauth2</artifactId>
+            <version>2.2.0.RELEASE</version>
         </dependency>
     </dependencies>
 
--- a/azure-spring-boot/pom.xml
+++ b/azure-spring-boot/pom.xml
@@ -25,9 +25,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.hibernate</groupId>
+            <groupId>org.hibernate.validator</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>6.0.4.Final</version>
             <optional>true</optional>
         </dependency>
 
--- a/azure-spring-boot/pom.xml
+++ b/azure-spring-boot/pom.xml
@@ -27,6 +27,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
+            <version>6.0.4.Final</version>
             <optional>true</optional>
         </dependency>
 
@@ -38,6 +39,7 @@
         <dependency>
             <groupId>org.json</groupId>
             <artifactId>json</artifactId>
+            <version>20171018</version>
         </dependency>
 
         <!--Azure active directory-->
--- a/azure-spring-boot/src/test/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilterPropertiesTest.java
+++ b/azure-spring-boot/src/test/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilterPropertiesTest.java
@@ -5,17 +5,14 @@
  */
 package com.microsoft.azure.spring.autoconfigure.aad;
 
-import javafx.print.Collation;
 import org.junit.Test;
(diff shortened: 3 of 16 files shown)


ZhijunZhao:
@snicoll It seems our starters can work well with SB 2.0 after a few minor changes. 

1. We're thinking about a better version name. How about 0.2.0-SB2? Any suggestion?

2. Should we release it and make it available in Spring Initalizr before Spring One?

snicoll:
> It seems our starters can work well with SB 2.0

Cool!

I remember that the active directory bits are much more easier with Spring Security 5 and perhaps some code on your side could be removed? I've pinged @rwinch on the original issue, perhaps he can share more details?

> We're thinking about a better version name

I wouldn't change the version structure (besides the version bump). First I would create a `0.2.x` branch (or whatever the current  version is) and I'd switch `master` to be Spring Boot 2 only and the branch to be Spring Boot 1 only. I guess you're going to continue improving the support for Spring Boot 1 users for some time still. `0.3.x` could be the range for Spring Boot 2 (or you could switch to `1.x`

> Should we release it and make it available in Spring Initalizr before Spring One?

I don't know. If you feel confident that it works well with Spring Boot 2, perhaps you could release a milestone of your own? There are a few issues in the PR though, I'll comment.

ZhijunZhao:
Add our PMs  @Matthew-Dong @selvasingh

azure-spring-boot-samples/azure-active-directory-spring-boot-sample/pom.xml
```diff
@@ -33,6 +33,7 @@
         <dependency>
             <groupId>org.springframework.security.oauth</groupId>
             <artifactId>spring-security-oauth2</artifactId>
+            <version>2.2.0.RELEASE</version>
```
snicoll: That part has been replaced in Spring Security 5 and rather than hardcoding the previous jar, I'd prefer if you port the feature to what Spring Security 5 now offers.

azure-spring-boot/pom.xml
```diff
@@ -27,6 +27,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
+            <version>6.0.4.Final</version>
```
snicoll: You shouldn't hardcode a version here. The version is still managed but the groupId has changed to `org.hibernate.validator`

azure-spring-boot/pom.xml
```diff
@@ -38,6 +39,7 @@
         <dependency>
             <groupId>org.json</groupId>
             <artifactId>json</artifactId>
+            <version>20171018</version>
```
snicoll: We've removed dependency management for json. Rather than keeping using it, perhaps you could move to something else (i.e. Jackson)

azure-spring-boot/src/test/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilterPropertiesTest.java
```diff
@@ -5,11 +5,20 @@
  */
 package com.microsoft.azure.spring.autoconfigure.aad;
 
+import javafx.print.Collation;
```
snicoll: That doesn't sound right.

rwinch:
> I remember that the active directory bits are much more easier with Spring Security 5 and perhaps some code on your side could be removed?

Spring Security 5 provides all you should need to do log in with OpenID Connect or an OAuth2 based log in. You can refer to https://docs.spring.io/spring-security/site/docs/current-SNAPSHOT/reference/htmlsingle/#jc-oauth2login for details

cc @jgrandja

sshogunn:
Hello,

Do you have any ideas when new version will be available?

ZhijunZhao:
@sshogunn We plan to resume the Spring Boot Support work in January 2018. New release could happen during that month or later.

Matthew-Dong:
@rwinch Hi Rob, we've been working on the integation of Spring Security 5. 

I'm wondering what would be the right "registrationId" for Azure AD?

jgrandja:
@Matthew-Dong The `registrationId` can be any arbitrary name you choose. There are no restrictions. You just need to ensure the `registrationId` is unique within a `ClientRegistrationRepository` if you have more than 1 client configured.

Matthew-Dong:
I see. Thanks for the answer, Joe. @jgrandja

sophiaso:
Upgraded to Spring Boot 2.0 in other PRs.

PR #252, opened 2018-01-25
Aad spring security 5

sophiaso:


azure-spring-boot-bom/pom.xml
```diff
@@ -46,6 +46,16 @@
     </scm>
```
ZhijunZhao: This BOM intended to provide dependency management for azure related packages only. For details, you can check issue #128.

Issue #259, opened 2018-02-02
PropertySettingUtil has a reference to a SNAPSHOT dependency


snicoll:
This [constant](https://github.com/Microsoft/azure-spring-boot/blob/6fac67e7ea1861165b9e2ae2981d5de73e1d91f2/azure-spring-boot/src/test/java/com/microsoft/azure/spring/autoconfigure/documentdb/PropertySettingUtil.java#L28) looks also suspicious.

Issue #258, opened 2018-02-02
Constants has a reference to a SNAPSHOT dependency


snicoll:
This [constant in `Constants`](https://github.com/Microsoft/azure-spring-boot/blob/6fac67e7ea1861165b9e2ae2981d5de73e1d91f2/azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/Constants.java#L11) looks highly suspicious.

sophiaso:
Fixed in #265 #263

PR #263, opened 2018-02-05
read version from properties

sophiaso:
@snicoll refs issues #258 #259

### DIFF
--- a/azure-spring-boot-parent/pom.xml
+++ b/azure-spring-boot-parent/pom.xml
@@ -79,6 +79,12 @@
     </dependencyManagement>
 
     <build>
+        <resources>
+            <resource>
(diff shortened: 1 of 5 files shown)


snicoll:
Why do you need to have access to this information?

yungez:
version info is part of usage data that we'd like to track.

snicoll:
It's not a blocker and feel free to merge this but have you considered using the manifest and set  the version for your root package instead? This [article is old but describes what I mean](https://www.javaworld.com/article/2074257/core-java/accessing-package-version-information-in-java-code.html)

sophiaso:
@snicoll Thank you, I have considered using manifest, but there is some test case relying on the version number and I need the version number in test phase, will double check the test case later.

PR #273, opened 2018-02-27
support build samples with gradle

yungez:


sophiaso:
Do we need to add java version property into the gradle build config file?

azure-spring-boot-samples/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+    version = '0.0.1-SNAPSHOT'
+}
+
+subprojects {
+
+    apply plugin: 'java'
+    apply plugin: 'org.springframework.boot'
+
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
```
yungez: @sophiaso  here's java version config: sourceCompatibility, targetCompatibility

PR #274, opened 2018-02-27
update gitingore for gradle build output

yungez:


.gitignore
```diff
@@ -51,6 +51,7 @@ __pycache__
 # gradle
 .gradle
 /build/
```
xscript: This line can be removed.

yungez: resolved. thanks.

PR #296, opened 2018-03-07
Fix sonar reported bugs

sophiaso:
## Summary
Fix some sonar reported bugs.

## Issue Type
- Bug fixing

## Starter Names
  - documentdb spring boot starter
  - azure cloud foundry service sample

## Additional Information

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
@@ -10,6 +10,8 @@ import com.microsoft.aad.adal4j.AuthenticationResult;
 import com.microsoft.aad.adal4j.ClientCredential;
 import com.microsoft.aad.adal4j.UserAssertion;
 
+import com.nimbusds.jose.JOSEException;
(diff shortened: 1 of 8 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
```diff
@@ -114,8 +118,12 @@ protected void doFilterInternal(
                 authentication.setAuthenticated(true);
                 log.info("Request token verification success. {}", authentication);
                 SecurityContextHolder.getContext().setAuthentication(authentication);
-            } catch (Throwable throwable) {
-                throw new RuntimeException(throwable);
+            } catch (MalformedURLException | ParseException | BadJOSEException | JOSEException ex) {
+                log.error("Failed to initialize UserPrincipal.", ex);
+                throw new ServletException(ex);
```
yungez: why not directly rethrow ex?

sophiaso: Because 4 new exception will be thrown if throw ex here, as these 4 exception are all thrown from initializing UserPrincipal, so wrapped into 1 exception here, the original exception will still be in the exception stacktrace.

Issue #231, opened 2017-11-15
NPE if VCAP_SERVICES is empty


jastev:
https://github.com/Microsoft/azure-spring-boot/blob/cb47d4ac857dfb82345c4827dbdd20f3600c5a57/azure-spring-boot/src/main/java/com/microsoft/azure/spring/cloundfoundry/environment/VcapProcessor.java#L72

You need to check here that names is not null, which it is if vcapServices is empty.  Otherwise, an NPE arises on the next line, when you access names.length().

PR #297, opened 2018-03-07
check null for names

sophiaso:
## Summary
Null check to avoid throwing NPE.

## Issue Type
- Bug fixing

## Starter Names
  - azure cloud foundry service sample

## Additional Information


sophiaso:
refs #231

Issue #307, opened 2018-03-23
`Azure Support` fails v1.5.10 spring-boot project


Eskibear:
## Environment
- Spring boot starter:
  - `Azure Support` in **Spring Initializr**
  
- OS Type: MacOS (I suppose the issue is cross-platform)
- Java version: 
  - 1.8
  

## Summary
Add `Azure Support` as a dependency, the spring-boot project fails to build.

## Reproduce steps
* open https://start.spring.io/
* select spring-boot version `1.5.10` 
* search and add `Azure Support` as the only dependency
* click `Generate Project` to download the zip file, and unzip it.
* in the folder, `mvn package` fails.
* even though `mvn package -DskipTests` can do the packaging, `mvn spring-boot:run` still fails with error.
 

## Expected Results
The project successfully runs.

## Actual Results
```
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.boot.maven.AbstractRunMojo$LaunchRunner.run(AbstractRunMojo.java:527)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NoClassDefFoundError: org/apache/commons/logging/LogFactory
	at org.springframework.boot.SpringApplication.<clinit>(SpringApplication.java:179)
	at com.example.demo.DemoApplication.main(DemoApplication.java:10)
	... 6 more
Caused by: java.lang.ClassNotFoundException: org.apache.commons.logging.LogFactory
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 8 more
```

sophiaso:
@Eskibear Thanks for reporting the issue, we have fixed the issue but not release yet. Please add the missing dependencies in above PR to your project for temp solution. If you are going to use Spring Boot Starter, then above missing dependencies are not necessary.

sophiaso:
2.0.x released, refs #323 

0.1.x not released

sophiaso:
[...]

PR #308, opened 2018-03-26
add missing dependencies which are required when using azure-spring-boot alone

sophiaso:
## Summary
Spring Boot application cannot start with azure-spring-boot dependency only.
These added dependencies will be provided by spring boot starters if any of spring boot starter is imported.

## Issue Type
- Bug fixing

## Starter Names
  - azure spring boot 

## Additional Information
refs #307

### DIFF
--- a/azure-spring-boot-parent/pom.xml
+++ b/azure-spring-boot-parent/pom.xml
@@ -80,6 +80,16 @@
                 <artifactId>json</artifactId>
                 <version>20171018</version>
             </dependency>
+            <dependency>
+                <groupId>commons-logging</groupId>
+                <artifactId>commons-logging</artifactId>
+                <version>1.2</version>
+            </dependency>
+            <dependency>
+                <groupId>org.yaml</groupId>
+                <artifactId>snakeyaml</artifactId>
+                <version>1.17</version>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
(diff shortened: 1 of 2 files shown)


azure-spring-boot/pom.xml
```diff
@@ -44,6 +44,14 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
         </dependency>
+        <dependency>
```
yungezz: why duplicate here? let's keep dependencies in parents only.

azure-spring-boot-parent/pom.xml
```diff
@@ -80,6 +80,16 @@
                 <artifactId>json</artifactId>
                 <version>20171018</version>
             </dependency>
+            <dependency>
+                <groupId>commons-logging</groupId>
+                <artifactId>commons-logging</artifactId>
+                <version>1.2</version>
```
yungezz: pls use properties for versions

sophiaso: Because all the versions in parent is written with explicit version instead of property. So just added as explicit version.

azure-spring-boot/pom.xml
sophiaso: In parent it's dependency management, we still need to add it as dependency without specifying version.

yungezz: got it. this is the bug caused the issue. any other newly introduced dependencies missing?

azure-spring-boot-parent/pom.xml
yungezz: pls help to update it as properties, when 2.0 release, i add same comments in some already merged pr

sophiaso: Moved version numbers to properties.

azure-spring-boot/pom.xml
sophiaso: This issue happens when user does not user any spring boot starter, by default, spring boot starter will add these dependencies.

PR #319, opened 2018-04-27
Replace deprecated hibernate validator dependency into javax.validation

rajadilipkolli:
## Summary
Hibernate validator is deprecated, we need to use javax.validation hence making changes as per that.


azure-spring-boot/pom.xml
```diff
@@ -24,12 +24,11 @@
...
-        <dependency>
-            <groupId>org.hibernate.validator</groupId>
-            <artifactId>hibernate-validator</artifactId>
-            <optional>true</optional>
-        </dependency>
-
+		<dependency>
+		    <groupId>javax.validation</groupId>
+		    <artifactId>validation-api</artifactId>
+		</dependency>
```
sophiaso: Hi @rajadilipkolli, thank you for sending PR to fix deprecated dependencies. Could you change the indention to align with other dependencies?

rajadilipkolli: @sophiaso , made changes to indention

rajadilipkolli: @sophiaso , made changes to indention

sophiaso: @rajadilipkolli  Thank you very much for the quick response.

PR #352, opened 2018-06-05
move graph APIs into separate class in order to be reused

sophiaso:
**Updated**: Some methods are moved from  `AADAuthenticationFilter.java` to `AzureADGraphClient.java`, in order to be reused later in spring-security-5 oauth2 update.

### Starter Names

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
@@ -64,8 +64,10 @@ public class AADAuthenticationFilter extends OncePerRequestFilter {
 
                 if (principal == null || graphApiToken == null || graphApiToken.isEmpty()) {
                     principal = new UserPrincipal(idToken, serviceEndpoints);
-                    graphApiToken =
-                            client.acquireTokenForGraphApi(idToken, principal.getClaim().toString()).getAccessToken();
+
+                    final String tenantId = principal.getClaim().toString();
+                    graphApiToken = client.acquireTokenForGraphApi(idToken, tenantId).getAccessToken();
+
                     principal.setUserGroups(client.getGroups(graphApiToken));
 
                     request.getSession().setAttribute(CURRENT_USER_PRINCIPAL, principal);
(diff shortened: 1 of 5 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
```diff
@@ -99,21 +60,21 @@ protected void doFilterInternal(
...
                 final ServiceEndpoints serviceEndpoints =
                         serviceEndpointsProp.getServiceEndpoints(aadAuthFilterProp.getEnvironment());
+                final AzureADGraphClient client = new AzureADGraphClient(aadAuthFilterProp, serviceEndpointsProp);
 
                 if (principal == null || graphApiToken == null || graphApiToken.isEmpty()) {
                     principal = new UserPrincipal(idToken, serviceEndpoints);
-                    graphApiToken = acquireTokenForGraphApi(
-                            idToken, principal.getClaim().toString(), serviceEndpoints).getAccessToken();
+                    graphApiToken =
+                            client.acquireTokenForGraphApi(idToken, principal.getClaim().toString()).getAccessToken();
```
Incarnation-p-lee: about line wrap, how about a local variable to hold `principal.getClaim().toString()` and pass to acquire for avoiding line wrap ?

sophiaso: Yes, I  think it's better and also better readability, will change.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -46,4 +80,73 @@ private static String getResponseStringFromConn(HttpURLConnection conn) throws I
...
+    }
+
+    private List<UserGroup> loadUserGroups(String graphApiToken) throws IOException {
+        final String responseInJson = getUserMembershipsV1(graphApiToken);
+        final List<UserGroup> lUserGroups = new ArrayList<>();
+        final ObjectMapper objectMapper = JacksonObjectMapperFactory.getInstance();
+        final JsonNode rootNode = objectMapper.readValue(responseInJson, JsonNode.class);
+        final JsonNode valuesNode = rootNode.get("value");
+        int i = 0;
+        while (valuesNode != null
```
Incarnation-p-lee: seems i only appears in loop, how about `for (int i; condition; i++)`

sophiaso: Thanks for the remind, will update.

sophiaso:
Updated according to the comments.

PR #365, opened 2018-06-25
Add auto configuration for spring-data-gremlin.

Incarnation-p-lee:
## Summary
* create Auto config when endpoint, port, username and password.
* create Auto config when Factory, Template and Converter.

## Issue Type

- New Feature

## Starter Names
  
  - spring data gremlin boot starter

## Additional Information

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinAutoConfiguration.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinAutoConfiguration.java
@@ -1,0 +1,96 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See LICENSE in the project root for
+ * license information.
+ */
+package com.microsoft.azure.spring.autoconfigure.gremlin;
+
+import com.microsoft.azure.telemetry.TelemetryData;
+import com.microsoft.azure.telemetry.TelemetryProxy;
+import com.microsoft.spring.data.gremlin.common.GremlinFactory;
+import com.microsoft.spring.data.gremlin.conversion.MappingGremlinConverter;
+import com.microsoft.spring.data.gremlin.mapping.GremlinMappingContext;
+import com.microsoft.spring.data.gremlin.query.GremlinTemplate;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.domain.EntityScanner;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.annotation.Persistent;
+import org.springframework.lang.NonNull;
+import org.springframework.util.ClassUtils;
+
+import java.util.HashMap;
+
+@Configuration
+@ConditionalOnClass({GremlinFactory.class, GremlinTemplate.class, MappingGremlinConverter.class})
+@ConditionalOnProperty(prefix = "gremlin", value = {"endpoint", "port", "username", "password"})
+@EnableConfigurationProperties(GremlinProperties.class)
+public class GremlinAutoConfiguration {
+
+    private final GremlinProperties properties;
+
+    private final TelemetryProxy telemetryProxy;
+
+    private final ApplicationContext applicationContext;
+
+    public GremlinAutoConfiguration(@NonNull GremlinProperties properties, @NonNull ApplicationContext context) {
+        this.properties = properties;
+        this.applicationContext = context;
+        this.telemetryProxy = new TelemetryProxy(properties.isTelemetryAllowed());
+    }
+
+    private void trackCustomEvent() {
+        final HashMap<String, String> customTelemetryProperties = new HashMap<>();
+        final String[] packageNames = this.getClass().getPackage().getName().split("\\.");
+
+        if (packageNames.length > 1) {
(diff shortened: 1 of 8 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinAutoConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+
+        telemetryProxy.trackEvent(ClassUtils.getUserClass(this.getClass()).getSimpleName(), customTelemetryProperties);
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    public GremlinFactory gremlinFactory() {
+        this.trackCustomEvent();
+
+        final String endpoint = this.properties.getEndpoint();
```
warrenzhu25: These 4 variable are unnecessary, in my opinion.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinAutoConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+    @Bean
+    @ConditionalOnMissingBean
+    public GremlinMappingContext gremlinMappingContext() {
+        try {
+            final GremlinMappingContext context = new GremlinMappingContext();
+
+            context.setInitialEntitySet(new EntityScanner(this.applicationContext).scan(Persistent.class));
+
+            return context;
+        } catch (ClassNotFoundException e) {
```
warrenzhu25: Which class is possibly not found? Could we use `ConditionalOnClass` to check this?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinProperties.java
```diff
@@ -0,0 +1,34 @@
...
+    @NotEmpty
+    private String port;
+
+    @NotEmpty
+    private String username;
+
+    @NotEmpty
+    private String password;
+
+    private boolean telemetryAllowed;
```
warrenzhu25: a default value for `telemetryAllowed`?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinAutoConfiguration.java
Incarnation-p-lee: In fact, there is a trade-off here, to take local variables or wrap the line. I prefer the first option.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinAutoConfiguration.java
yaweiw: sounds a good point.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinProperties.java
Incarnation-p-lee: should be `true` here, will fix this.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinAutoConfiguration.java
yaweiw: I also don't understand why the local vars are needed here. can you please elaborate?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinAutoConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+            customTelemetryProperties.put(TelemetryData.SERVICE_NAME, packageNames[packageNames.length - 1]);
+        }
+
+        telemetryProxy.trackEvent(ClassUtils.getUserClass(this.getClass()).getSimpleName(), customTelemetryProperties);
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    public GremlinFactory gremlinFactory() {
+        this.trackCustomEvent();
```
warrenzhu25: If user provide his own `GremlinFactory`, should we track event?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/gremlin/GremlinAutoConfiguration.java
Incarnation-p-lee: `scan` may throw `NotFoundException`, here we do not reference to `Conditional` because we treat it as somewhat Fatal error. There is no persistent class for database, the user may need to check the code.

warrenzhu25: How other spring data project handle such scenario? Relying on Exception seems inelegant.

Incarnation-p-lee: Almost the same, in our `cosmosdb` it catch the exception, in `mongodb` it just throws the exception. Both are relying on Exceptions.

yaweiw: I think using exception to manipulate app logic isn't the right practice. But since in this case you want to notify the user of failure, it makes sense to rethrow.

Incarnation-p-lee: agree, just trade off here if there is no better choice.

PR #390, opened 2018-08-01
Update maven dependency for active directory starter readme

patrick246:
## Summary
The code sample below doesn't work in version 2.0.2, as the `OAuth2UserService` isn't yet provided in this version. A bump to 2.0.4 fixes that problem.

## Issue Type
- Bug fixing

## Starter Names

## Additional Information
In the future, this section could probably extended by including the maven bom, but this is a quick fix, that keeps the documentation accurate


sophiaso:
@patrick246  @annikahanna Thanks for the effort.

PR #407, opened 2018-09-02
enabled MSI support in KeyVault module

jalanubha:
## Summary
Enabled Managed Service Identity support in the KeyVault module

## Issue Type
- New Feature

## Starter Names

## Additional Information


azure-spring-boot/pom.xml
```diff
@@ -139,6 +139,11 @@
             <artifactId>adal4j</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>azure-client-authentication</artifactId>
+            <version>1.6.1</version>
```
sophiaso: @jalanubha Thanks for the pull request, could you manage version in [BOM file](https://github.com/Microsoft/azure-spring-boot/blob/master/azure-spring-boot-bom/pom.xml) and write dependency without version here? 

If `azure-client-authentication` may not  be required by all auto-configuration classes, please add `<optional>true</optional>`.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/Constants.java
```diff
@@ -12,6 +12,8 @@
     public static final String AZURE_KEYVAULT_USER_AGENT = "spring-boot-starter/" + PropertyLoader.getProjectVersion();
     public static final String AZURE_CLIENTID = "azure.keyvault.client-id";
     public static final String AZURE_CLIENTKEY = "azure.keyvault.client-key";
+    public static final String AZURE_MSI_ENDPOINT = "MSI_ENDPOINT";
+    public static final String AZURE_MSI_SECRET = "MSI_SECRET";
```
sophiaso: As the property here is Spring properties, better to align the properties name with other properties? e.g., `azure.keyvault.msi-endpoint` and `azure.keyvault.msi-secret`?

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelper.java
```diff
@@ -34,18 +36,25 @@ public KeyVaultEnvironmentPostProcessorHelper(final ConfigurableEnvironment envi
...
                 .orElse(1800000L);
         final long timeAcquiringTimeoutInSeconds = this.environment.getProperty(
-                Constants.AZURE_TOKEN_ACQUIRE_TIMEOUT_IN_SECONDS, Long.class, 60L);
-
-        final ServiceClientCredentials credentials = new AzureKeyVaultCredential(clientId, clientKey,
+            Constants.AZURE_TOKEN_ACQUIRE_TIMEOUT_IN_SECONDS, Long.class, 60L);
+        final ServiceClientCredentials credentials;
+
+        if (this.environment.containsProperty(Constants.AZURE_MSI_ENDPOINT)
+            && this.environment.containsProperty(Constants.AZURE_MSI_SECRET)) {
```
Incarnation-p-lee: looks we need to add test for this part

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/Constants.java
jalanubha: How about I use these string constants directly in the destination file "KeyVaultEnvironmentPostProcessorHelper.java"? as MSI_ENDPOINT and MSI_SECRET are the default environment variables set in the App Service environment

nickwalkmsft:
I'd really like to see this happen - is there anything needed to push this forward?

jalanubha:
@sophiaso Added 1 more test case. Please review.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelper.java
```diff
@@ -41,11 +41,11 @@ public void addKeyVaultPropertySource() {
...
                 .map(Long::valueOf).orElse(Constants.DEFAULT_REFRESH_INTERVAL_MS);
         final long timeAcquiringTimeoutInSeconds = this.environment.getProperty(
-            Constants.AZURE_TOKEN_ACQUIRE_TIMEOUT_IN_SECONDS, Long.class, 60L);
+            Constants.AZURE_TOKEN_ACQUIRE_TIMEOUT_IN_SECONDS, Long.class, Constants.TOKEN_ACQUIRE_TIMEOUT_SECS);
         final ServiceClientCredentials credentials;
 
-        if (this.environment.containsProperty(Constants.AZURE_MSI_ENDPOINT)
-            && this.environment.containsProperty(Constants.AZURE_MSI_SECRET)) {
+        if (this.environment.containsProperty("MSI_ENDPOINT")
+            && this.environment.containsProperty("MSI_SECRET")) {
```
simholte: Shouldn't these be properties that are passed in by the client?  The way it's written right now I could configure it to use a client-id and client-secret and if the environment variable MSI_ENDPOINT and MSI_SECRET exist it would use MSI instead.  Seems misleading to me

simholte:
When I build the PR and use the library I get a null pointer exception when it calls trackCustomEvent, specifically when it flushes the telemetry data

Incarnation-p-lee:
@simholte
You mean the telemetryclient created and then met NPE when sending dat, right ?

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelper.java
jalanubha: No, these properties aren't passed by the client. When you host the app on Azure App service environment, it will be picked up from there if MSI is enabled in the App service.

Yes, that is right, the way it is written is if you have MSI enabled, always use that else fallback to the usual client secret authentication. That is the expected behavior as per my understanding.

simholte:
@Incarnation-p-lee 
That's correct.

Incarnation-p-lee:
@dhaval24 
It seems the AI core part, Cloud you please help to take a look ?
As the TelemetryClient instance created already, looks nothing we can do from this repo.

dhaval24:
@simholte could you please provide the stack trace of the exception your facing. It will help me better understand. I haven't seen any issue with ApplicationInsights Java SDK in general.

Incarnation-p-lee:
@dhaval24 
Hi is there any update about this NPE, from the description in above, I think it may be reproduced with this PR of the repo. Cloud you please help to have a try ?

azure-spring-boot/src/test/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelperUnitTest.java
```diff
@@ -0,0 +1,62 @@
...
+    public void testMSIAuthPropertiesNotInitialized() { 
+        assertFalse("Environment should not contain MSI_ENDPOINT when not on Azure App Service environment",
+        environment.containsProperty("MSI_ENDPOINT"));
+
+        assertFalse("Environment should not contain MSI_SECRET when not on Azure App Service environment",
+        environment.containsProperty("MSI_SECRET"));
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void testMSIAuthentication() {
```
elhoo: I can't really see what this test verifies except that it except a RuntimeException to be thrown?

sophiaso: Agreed, thanks, cc @jalanubha

dhaval24:
Pan, thanks for reminding. I have not got a chance to look at the issue. I will try to take a look this weekend. In the meanwhile can we get SDKLogs? We can enable it by adding <SDKLogger /> tag in the applicationinsights.xml file and rebuild.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelper.java
```diff
@@ -34,17 +36,24 @@ public KeyVaultEnvironmentPostProcessorHelper(final ConfigurableEnvironment envi
...
         final long timeAcquiringTimeoutInSeconds = this.environment.getProperty(
-                Constants.AZURE_TOKEN_ACQUIRE_TIMEOUT_IN_SECONDS, Long.class, Constants.TOKEN_ACQUIRE_TIMEOUT_SECS);
-
-        final ServiceClientCredentials credentials = new AzureKeyVaultCredential(clientId, clientKey,
+            Constants.AZURE_TOKEN_ACQUIRE_TIMEOUT_IN_SECONDS, Long.class, Constants.TOKEN_ACQUIRE_TIMEOUT_SECS);
+        final ServiceClientCredentials credentials;
+
+        if (this.environment.containsProperty("MSI_ENDPOINT")
+            && this.environment.containsProperty("MSI_SECRET")) {
+            credentials = new AppServiceMSICredentials(AzureEnvironment.AZURE);
```
sophiaso: This constructor `AppServiceMSICredentials(AzureEnvironment environment)` is going to load the `MSI_ENDPOINT` and `MSI_SECRET` from `System environment variable`, while the `environment.containsProperty` only promises the properties can be found in the environment, but cannot promise the `System` contains those variables, which may cause potential issues. May need enhancement in another pull request.

nickwalkmsft:
👍 At what point does this get published and become consumable via start.spring.io?

nickwalkmsft:
Ping on the above question...

sophiaso:
Hi @nickwalkmsft, a  new  version may be released within this week or early next week.

sophiaso:
Version `2.0.7` has been released.

PR #410, opened 2018-09-04
replace json with jackson (fix #233)

### LINKED ISSUE #233: Use Jaskson instead of json

sophiaso:
## Summary
Spring Boot provides dependency management for jackson, but not org.json, migrate to jackson.

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/cloundfoundry/environment/VcapProcessor.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/cloundfoundry/environment/VcapProcessor.java
@@ -5,9 +5,9 @@
  */
 package com.microsoft.azure.spring.cloundfoundry.environment;
 
-import org.json.JSONArray;
-import org.json.JSONException;
-import org.json.JSONObject;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationFeature;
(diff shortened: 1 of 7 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/cloundfoundry/environment/VcapProcessor.java
```diff
@@ -76,27 +67,33 @@ public VcapResult parse(String vcapServices) {
...
-                    for (int i = 0; i < names.length(); i++) {
-                        final String name = (String) names.get(i);
                         if (name.startsWith(AZURE) || USER_PROVIDED.equals(name)) {
-                            final JSONArray azureService = json.getJSONArray(name);
-                            final int numElements = azureService.length();
-                            for (int index = 0; index < numElements; index++) {
-                                final VcapPojo pojo = parseService(name, azureService,
-                                        vcapServices, index);
-                                
+                            final List<VcapServiceConfig> azureServices = serviceEntry.getValue();
```
Incarnation-p-lee: I am not sure here we can change to this ?
```java
results.addAll(azureServices.stream().map(s -> parseService(name, s, vcapServices).filter(p -> p != null))
```

sophiaso: Updated with lambda expression.

PR #414, opened 2018-09-05
refactoring keyvault to allow different types of data stored

sophiaso:
## Summary
Migrate key vault from Spring Cloud Azure with some changes, to allow different implementations. 

Test is not updated, will be in another PR.


azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/secrets/KeyVaultSecretTemplate.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public KeyVaultSecretTemplate(String clientId, String clientSecret, long refreshIntervalMS) {
+        super(clientId, clientSecret, refreshIntervalMS);
+    }
+
+    @Override
+    public String getSecret(String keyVaultName, String secretName) {
+        final Pair<String, String> keyVaultAndSecret = Pair.of(keyVaultName, secretName);
+        if (useCache) {
+            clearCacheAfterInterval();
```
warrenzhu25: I suggest not to clear cache, what we really need is update new values. Just leave old values there.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/secrets/KeyVaultSecretTemplate.java
```diff
@@ -0,0 +1,94 @@
...
+        final String keyVaultName = keyVaultAndSecret.getLeft();
+
+        final SecretBundle bundle = this.keyVaultClientCreator.apply(keyVaultName)
+                .getSecret(buildKeyVaultUrl(keyVaultName), keyVaultAndSecret.getRight());
+
+        return bundle == null ? null : bundle.value();
+    }
+
+    private void clearCacheAfterInterval() {
+        if (refreshIntervalMS <= 0) {
```
warrenzhu25: If we want to refresh periodically, I suggest to use separate scheduled thread pool to do this. This will improve latency when user really want to use such value.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
```diff
@@ -3,102 +3,15 @@
...
-                        .replaceFirst(this.vaultUri + "/secrets/", ""), secret.id());
-            }
+    void setRefreshInterval(long refreshIntervalMilliSecs);
 
-            this.lastUpdateTime.set(System.currentTimeMillis());
-        } finally {
-            this.rwLock.writeLock().unlock();
-        }
-    }
+    void setAllowTelemetry(boolean allowTelemetry);
```
warrenzhu25: This shouldn't be part of interface. If others want to provide another implementation. Why should we bother with this?

warrenzhu25: Also for other 2 methods.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/AbstractKeyVaultTemplate.java
```diff
@@ -0,0 +1,69 @@
...
+
+import static com.microsoft.azure.keyvault.spring.Constants.DEFAULT_REFRESH_INTERVAL_MS;
+
+public abstract class AbstractKeyVaultTemplate implements KeyVaultOperation {
+    private static final String KEY_VAULT_URL = "https://%s.vault.azure.net/";
+
+    protected final AzureKeyVaultCredential credential;
+    protected Function<String, KeyVaultClient> keyVaultClientCreator = keyVaultClientCreator();
+    private boolean allowTelemetry = false;
+    protected long refreshIntervalMS = DEFAULT_REFRESH_INTERVAL_MS;
```
warrenzhu25: I suggest to have kind of option to disable periodic refresh. I would say in most cases, secret won't change. Maybe if this property is not provided, no refresh.

sophiaso: With default refresh interval(-1), it will not be refreshed.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
sophiaso: Moved to abstract template and now only keep the KeyVaultOperation, and we can provide more interfaces if we decide to support certificates etc. in KeyVault.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/secrets/KeyVaultSecretTemplate.java
sophiaso: Wonder is it too heavy to introduce thread pool here.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/secrets/KeyVaultSecretTemplate.java
```diff
@@ -0,0 +1,98 @@
...
+        return bundle == null ? null : bundle.value();
+    }
+
+    private void scheduleRefresh(long refreshIntervalMS) {
+        if (refreshIntervalMS <= 0) {
+            // Do not refresh
+            return;
+        }
+
+        final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
```
warrenzhu25: How about making this as class field? otherwise this might be garbage collected.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/secrets/KeyVaultSecretTemplate.java
sophiaso: Even though we refresh periodically, but the secrets still have to be queried, there is no existing way to check updates only.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
sophiaso: Moved to the abstract template.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/AbstractKeyVaultTemplate.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.function.Function;
+
+import static com.microsoft.azure.keyvault.spring.Constants.DEFAULT_REFRESH_INTERVAL_MS;
+
+public abstract class AbstractKeyVaultTemplate implements KeyVaultOperation {
+    private static final String KEY_VAULT_URL = "https://%s.vault.azure.net/";
+
+    protected final AzureKeyVaultCredential credential;
+    protected Function<String, KeyVaultClient> keyVaultClientCreator = keyVaultClientCreator();
```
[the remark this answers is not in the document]
sophiaso: Added final.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/secrets/KeyVaultSecretTemplate.java
```diff
@@ -0,0 +1,98 @@
...
+        final Runnable refresher = new Runnable() {
+            @Override
+            public void run() {
+                final Set<String> keyVaultNames = secretNamesByKeyVault.keySet();
+                secretsByKeyVault.clear();
+
+                keyVaultNames.stream().forEach(keyVaultName -> {
+                    final Set<String> secretNames = fetchSecretNames(keyVaultName);
```
warrenzhu25: I suggest to rewrite this forEach logic. Generally, forEach doesn't have another forEach inside.

warrenzhu25: I think you should update `secretNamesByKeyVault` with new `secretNames` here

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/secrets/KeyVaultSecretTemplate.java
```diff
@@ -0,0 +1,98 @@
...
+
+                    secretNames.stream().forEach(secretName -> {
+                        final Pair<String, String> keyVaultSecrect = Pair.of(keyVaultName, secretName);
+                        final String secretValue = fetchSecret(keyVaultSecrect);
```
warrenzhu25: How about parallel `fetchSecret`? It's a remote call, calling it one by one will be too slow.

chenrujun:
I'll close this PR because it's not updated for more than 1 year.

Issue #321, opened 2018-04-27
Azure AD implementation not compatible with session sharing due to missing Serializable


chrisgahlert:
## Environment
- Spring boot starter:
  - active directory spring boot starter
  
## Summary
After validating the Bearer token in `AADAuthenticationFilter`, a new `UserPrincipal` object is created and saved to the session.

As the `UserPrincipal` class is not implementing the `Serializable` interface. It also doesn't support being serialized and deserialized to/from JSON, as the __JSON Web Key__ classes don't support this.

For this reason, it is impossible to enable session sharing between multiple application instances (e.g. via spring-session).

## Solution
Either have the `UserPrincipal` class implement the `Serializable` interface or don't persist this object in the session. Instead one could persist the properties needed (Id, Claims, Groups, ...) to create the `UserPrincipal` inside the `AADAuthenticationFilter`.

mapster:
Can we expect this to be fixed anytime soon?

sophiaso:
A new version [2.0.7](https://github.com/Microsoft/azure-spring-boot/releases/tag/2.0.7) has been released to support serializable UserPrincipal. Close issue.

PR #427, opened 2018-10-09
Refactor aad to support serializable UserPrincipal

sophiaso:
UserPrincipal was not serializable caused by 1) not implementing serializable 2) includes non-serializable variable JWKSet, as the JWKSet value can be loaded by user from public endpoint, it's removed from the UserPrincipal.

refs #321

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/UserPrincipalManager.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/UserPrincipalManager.java
@@ -27,9 +27,9 @@ import java.text.ParseException;
 public class UserPrincipalManager {
     private static final Logger LOG = LoggerFactory.getLogger(UserPrincipalManager.class);
 
-    private ServiceEndpoints serviceEndpoints;
-    private ConfigurableJWTProcessor<SecurityContext> validator;
-    private ResourceRetriever resourceRetriever;
+    private final ServiceEndpoints serviceEndpoints;
+    private final ConfigurableJWTProcessor<SecurityContext> validator;
+    private final ResourceRetriever resourceRetriever;
 
     public UserPrincipalManager(ServiceEndpoints serviceEndpoints, ResourceRetriever resourceRetriever) {
         this.serviceEndpoints = serviceEndpoints;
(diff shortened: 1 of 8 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/UserPrincipalManager.java
```diff
@@ -0,0 +1,77 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.text.ParseException;
+
+public class UserPrincipalManager {
+    private static final Logger LOG = LoggerFactory.getLogger(UserPrincipalManager.class);
+
+    private ServiceEndpoints serviceEndpoints;
+    private ConfigurableJWTProcessor<SecurityContext> validator;
+    private ResourceRetriever resourceRetriever;
```
Incarnation-p-lee: Do we need to add final for these 3 fields ?

azure-spring-boot/src/test/java/com/microsoft/azure/spring/autoconfigure/aad/UserPrincipalTest.java
```diff
@@ -75,4 +82,34 @@ public void getGroups() throws Exception {
...
     }
+
+    @Test
+    public void userPrinciplaIsSerializable() throws ParseException, IOException, ClassNotFoundException {
+        final File tmpOutputFile = File.createTempFile("test-user-principal", "txt");
+
+        try (final FileOutputStream fileOutputStream = new FileOutputStream(tmpOutputFile);
+             final ObjectOutputStream objectOutputStream = new ObjectOutputStream(fileOutputStream);
+             final FileInputStream fileInputStream = new FileInputStream(tmpOutputFile);
+             final ObjectInputStream objectInputStream = new ObjectInputStream(fileInputStream);){
```
Incarnation-p-lee: seems not aligned ?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/UserPrincipalManager.java
sophiaso: final added.

Issue #431, opened 2018-10-10
Upgrade azure storage sdk to v10


sophiaso:
- [x] Upgrade sdk to azure-storage-blob v10, #434 

- [x] Update azure-storage-spring-boot-starter README

- [x] Update storage spring boot starter sample

SuhanKoh:
Hi @sophiaso, is there an ETA for azure storage sdk v10 update?

sophiaso:
Hi @SuhanKoh, probably being released within two weeks.

sophiaso:
Hi @SuhanKoh, starter with azure storage sdk v10 has been released, [maven link](https://search.maven.org/search?q=g:com.microsoft.azure%20AND%20a:azure-storage-spring-boot-starter)

PR #434, opened 2018-10-11
upgrade storage sdk to v10

sophiaso:
refs #431

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/storage/StorageAutoConfiguration.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/storage/StorageAutoConfiguration.java
@@ -61,12 +61,7 @@ public class StorageAutoConfiguration {
     }
 
     private PipelineOptions buildOptions(PipelineOptions fromOptions) {
-        final PipelineOptions pipelineOptions;
-        if (fromOptions != null) {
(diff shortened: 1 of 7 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/storage/StorageAutoConfiguration.java
```diff
@@ -39,26 +42,42 @@ public StorageAutoConfiguration(StorageProperties properties) {
...
-    private CloudStorageAccount createCloudStorageAccount() throws URISyntaxException, InvalidKeyException {
-        LOG.debug("createCloudStorageAccount called");
-        return CloudStorageAccount.parse(properties.getConnectionString());
+    private PipelineOptions buildOptions(PipelineOptions fromOptions) {
+        final PipelineOptions pipelineOptions;
+        if (fromOptions != null) {
+            pipelineOptions = fromOptions;
+        } else {
+            pipelineOptions = new PipelineOptions();
+        }
```
Incarnation-p-lee: As the if-else is simple, suggest to convert to `option = fromOption != null ? fromOPtion : new option()`

sophiaso: Updated expression.

Issue #464, opened 2018-11-16
Integration AAD B2C with spring security


Incarnation-p-lee:
The feature implementation will be merge to one development branch [aad-b2c-integration](https://github.com/Microsoft/azure-spring-boot/tree/aad-b2c-integration).

Feature check list:
* OpenId
  * [x] Sign in/up/out
  * [x] Password reset
  * [x] Profile edit
  * [x] Forgot password (workaround for AAD B2C sign in page)
* Issues
  * [ ] Support both relative and absolute URL from configuration
  * [ ] Investigate the best practice when obtain JWT keyset
  * [x] #481 
* Investigate
  * [x] formLogin
  * [x] openidLogin
  * [ ] session stateless with JWT
     *  Normal way of jQuery or XMLHttpRequest with id_token need cors support, but b2c call it unsupported. [CORS in B2C](https://stackoverflow.com/questions/34200409/cors-issue-when-getting-a-token-in-azure-ad-b2c-implict-flow)
    * Need investigate other way when request to backend.

Incarnation-p-lee:
current aad use fixed algorithm in jwt parser, not sure if may related to #454, may be double check that part.

SteefB:
What is the current state of the integration? When can we expect to use B2C with spring security? Is there anything I can do to speed up the process? Thanks!

Incarnation-p-lee:
@SteefB 
Thanks for reminding. We have [release milestone](https://search.maven.org/search?q=g:com.microsoft.azure%20AND%20a:azure-active-directory-b2c-spring-boot-starter) for this with spring boot 2.1.x.

You can follow the sample [here](https://github.com/Microsoft/azure-spring-boot/tree/master/azure-spring-boot-samples/azure-active-directory-b2c-oidc-spring-boot-sample).

Incarnation-p-lee:
@SteefB 
Or you can leverage snapshot version.

```
<repositories>
  <repository>
    <id>nexus-snapshots</id>
    <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
    <snapshots>
      <enabled>true</enabled>
      <updatePolicy>always</updatePolicy>
    </snapshots>
  </repository>
</repositories>
```

SteefB:
@Incarnation-p-lee Thanks for the information! We'll check it out asap. 
Do you have estimation on when this will available in an official release? 
Thanks a lot!

erviszyka:
Hello all,
[...]

PR #465, opened 2018-11-16
Add b2c authentication properties with auto-config, #464.

Incarnation-p-lee:
* Inject properties to auto-configuration.
* Add test for bean and properties.

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthenticationProperties.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthenticationProperties.java
@@ -1,0 +1,91 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See LICENSE in the project root for
+ * license information.
+ */
+package com.microsoft.azure.spring.autoconfigure.b2c;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import org.hibernate.validator.constraints.URL;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+import javax.validation.constraints.NotBlank;
+
+@Getter
+@Setter
+@Validated
+@ConfigurationProperties(prefix = AADB2CAuthenticationProperties.PREFIX)
+public class AADB2CAuthenticationProperties {
+
+    public static final String PREFIX = "azure.activedirectory.b2c";
+
+    private static final String POLICIES = "policies";
+
+    private static final String SIGN_UP_OR_SIGN_IN = "sign-up-or-sign-in";
+
+    /**
+     * We do not use ${@link String#format(String, Object...)}
+     * as it's not real constant, which cannot be referenced in annotation.
+     */
+    private static final String POLICY_SIGN_UP_OR_SIGN_IN = POLICIES + "." + SIGN_UP_OR_SIGN_IN;
+
+    public static final String POLICY_SIGN_UP_OR_SIGN_IN_NAME = POLICY_SIGN_UP_OR_SIGN_IN + ".name";
+
+    public static final String POLICY_SIGN_UP_OR_SIGN_IN_REDIRECT_URL = POLICY_SIGN_UP_OR_SIGN_IN + ".redirect-url";
+
+    /**
+     * The name of the b2c tenant that created.
+     */
+    @NotBlank(message = "tenant name should not be blank")
+    private String tenant;
+
+    /**
+     * The application ID that registered under b2c tenant.
+     */
+    @NotBlank(message = "client ID should not be blank")
+    private String clientId;
+
+    /**
+     * The all polices that created under b2c tenant
+     */
+    @JsonProperty(POLICIES)
+    private Policies policies = new Policies();
+
+    @Getter
+    @Validated
+    @NoArgsConstructor(access = AccessLevel.PROTECTED)
+    public static class Policies {
+
+        /**
+         * The sign-up-or-sign-in policies that created under b2c tenant.
+         */
+        @JsonProperty(SIGN_UP_OR_SIGN_IN)
+        private Policy signUpOrSignIn = new Policy();
+
+        // TODO(pan): will add more policies like sign-in, sign-up, profile-editing and password-reset.
+    }
+
+    @Getter
+    @Setter
+    @Validated
+    @NoArgsConstructor(access = AccessLevel.PROTECTED)
+    public static class Policy {
+
+        /**
+         * The name of policy that created under b2c tenant.
+         */
+        @NotBlank(message = "policy name should not be blank")
+        private String name;
+
+        /**
+         * The redirect URL that configured under
+         */
+        @URL(message = "redirect URL should be valid")
+        private String redirectUrl;
+    }
+}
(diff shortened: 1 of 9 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthenticationProperties.java
```diff
@@ -0,0 +1,91 @@
...
+@Setter
+@Validated
+@ConfigurationProperties(prefix = AADB2CAuthenticationProperties.PREFIX)
+public class AADB2CAuthenticationProperties {
+
+    public static final String PREFIX = "azure.activedirectory.b2c";
+
+    private static final String POLICIES = "policies";
+
+    private static final String SIGN_UP_OR_SIGN_IN = "sign-up-or-sign-in";
```
warrenzhu25: Do we really need to distinguish sign in or up here?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthenticationProperties.java
```diff
@@ -0,0 +1,91 @@
...
+    public static class Policy {
+
+        /**
+         * The name of policy that created under b2c tenant.
+         */
+        @NotBlank(message = "policy name should not be blank")
+        private String name;
+
+        /**
+         * The redirect URL that configured under
```
warrenzhu25: under what?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthenticationProperties.java
Incarnation-p-lee: Yes, the naming is from portal, as there are sign-in and sign-up policies.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthenticationProperties.java
Incarnation-p-lee: Oh, no, break by yitao...

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthenticationProperties.java
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+import javax.validation.constraints.NotBlank;
+
+@Getter
+@Setter
+@Validated
+@ConfigurationProperties(prefix = AADB2CAuthenticationProperties.PREFIX)
+public class AADB2CAuthenticationProperties {
```
xscript: Do we really need `Authentication` in the class name?

Incarnation-p-lee: In previous, I prefer to remove this word from classname. But security contains both authentication and authorize, so I keep this word in case we may need authorize properties in future. If we can merge these 2 parts, then we can remove this words safely.

Incarnation-p-lee:
FYI
```
azure:
  activedirectory:
    b2c:
      tenant: xxxxx
      client-Id: xxxxxx
      policies:
        sign-up-or-sign-in:
          name: xxxx
          redirect-url: xxxx
        profile-editing:
          name: xxxx
          redirect-url:
        password-rest: xxxx
          name: xxxx
          redirect-url: xxx
```

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CProperties.java
```diff
@@ -0,0 +1,91 @@
...
+    private String tenant;
+
+    /**
+     * The application ID that registered under b2c tenant.
+     */
+    @NotBlank(message = "client ID should not be blank")
+    private String clientId;
+
+    /**
+     * The all polices that created under b2c tenant
```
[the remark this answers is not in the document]
Incarnation-p-lee: Yes, policies is created for each application but B2C organize the `policies` and `applications` in same level, and all the applications share all policies, as well as users.
Base on that I do not want to show any hint that `policies` is a part of `Application`, and only mention `tenant` here.

PS: will update comments for period.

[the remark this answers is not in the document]
xscript: So it means policies are shared among different applications. Didn't notice that before.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CProperties.java
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * The application ID that registered under b2c tenant.
+     */
+    @NotBlank(message = "client ID should not be blank")
+    private String clientId;
+
+    /**
+     * The all polices that created under b2c tenant.
+     */
+    @JsonProperty(POLICIES)
```
sophiaso: The constant `POLICIES ` has same value with the property name `policies`, the `@JsonProperty` annotation may not be necessary?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAutoConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+@Configuration
+@ConditionalOnWebApplication
+@ConditionalOnProperty(prefix = AADB2CProperties.PREFIX, value = {
+        "tenant",
+        "client-id",
+        POLICY_SIGN_UP_OR_SIGN_IN_NAME,
+        POLICY_SIGN_UP_OR_SIGN_IN_REDIRECT_URL,
+})
+@EnableConfigurationProperties(AADB2CProperties.class)
+public class AADB2CAutoConfiguration {
```
warrenzhu25: In Java naming convention, this might be `AadB2c`.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CProperties.java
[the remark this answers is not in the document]
Incarnation-p-lee: Yes, also surprise me when confirmed this with service team.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAutoConfiguration.java
warrenzhu25: Does this config need some condition on class?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CProperties.java
Incarnation-p-lee: Use constant `POLICIES` here for less possibility of typo, as it may depends on the String literal one line 35.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAutoConfiguration.java
Incarnation-p-lee: For noun, Yes. But here is abbreviation, like `java.net.URI`.

PR #467, opened 2018-11-19
Add aad b2c entry point, #464

Incarnation-p-lee:
* Add b2c entry point, 
* Add bean test for entry point.

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
@@ -88,6 +88,8 @@ public class AADB2CURL {
         final String endpoint = format(API_PATTERN, properties.getTenant(), properties.getTenant(), API_TYPE_AUTHORIZE);
         final AADB2CProperties.Policy policy = properties.getPolicies().getSignUpOrSignIn();
         final List<String> parameters = Arrays.asList(
+                // Each element is one parameter of redirect URL that take 'property=value' format.
+                // For example, 'response_mode=query' and 'response_type=code+id_token'.
                 format(PARAMETER_PATTERN, CLIENT_ID, properties.getClientId()),
                 format(PARAMETER_PATTERN, REDIRECT_URL, getEncodedURL(policy.getRedirectURI())),
                 format(PARAMETER_PATTERN, RESPONSE_MODE, RESPONSE_MODE_QUERY),
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
@@ -13,8 +13,6 @@ import org.springframework.util.Assert;
 import javax.servlet.http.HttpServletRequest;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
-import java.util.Arrays;
-import java.util.List;
 import java.util.UUID;
 
 import static java.lang.String.format;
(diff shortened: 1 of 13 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
```diff
@@ -0,0 +1,103 @@
...
+    public static String getOpenIdSignUpOrSignInUrl(@NonNull AADB2CProperties properties, String requestURL) {
+        Assert.hasText(requestURL, "requestURL should have text.");
+
+        final String endpoint = format(API_PATTERN, properties.getTenant(), properties.getTenant(), API_TYPE_AUTHORIZE);
+        final AADB2CProperties.Policy policy = properties.getPolicies().getSignUpOrSignIn();
+        final List<String> parameters = Arrays.asList(
+                format(PARAMETER_PATTERN, CLIENT_ID, properties.getClientId()),
+                format(PARAMETER_PATTERN, REDIRECT_URL, getEncodedURL(policy.getRedirectURI())),
+                format(PARAMETER_PATTERN, RESPONSE_MODE, RESPONSE_MODE_QUERY),
+                format(PARAMETER_PATTERN, RESPONSE_TYPE, format("%s+%s", RESPONSE_TYPE_CODE, RESPONSE_TYPE_ID_TOKEN)),
```
xscript: Will we allow customization of response type?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param properties of ${@link AADB2CProperties}
+     * @param requestURL from ${@link HttpServletRequest} that user attempt to access.
+     * @return
+     */
+    public static String getOpenIdSignUpOrSignInUrl(@NonNull AADB2CProperties properties, String requestURL) {
+        Assert.hasText(requestURL, "requestURL should have text.");
+
+        final String endpoint = format(API_PATTERN, properties.getTenant(), properties.getTenant(), API_TYPE_AUTHORIZE);
+        final AADB2CProperties.Policy policy = properties.getPolicies().getSignUpOrSignIn();
+        final List<String> parameters = Arrays.asList(
```
xscript: Can we use a big fat well-formatted string pattern here? I don't see why we need so many small strings.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
Incarnation-p-lee: I think it is unnecessary here, as the response type is `id_token` or `id_token+code`, we will wrapper these two field in `UserPrincipal` and the customer application can obtain want they want then.
Unless there are some customers require this configuration.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
Incarnation-p-lee: as variable naming hint, it is `parameters` for redirect URL, for each line is one parameter like `%s=%s`, or I may add some comments for this.
The big format may looks like
`"%s=%s&%s=%s&%s=%s&%s=%s&%s=%s&%s=%s&%s=%s&%s=%s`, If we leverage this one,  we need to take about 16 parameters to the String.format, I don't think that is a good idea.

xscript: Leave the choice to you  since this is just personal preference. 😄

Incarnation-p-lee: Thanks, update to big pattern for consideration of more straight forward, as well as readability.

PR #468, opened 2018-11-19
Add b2c logout success handler, #464.

Incarnation-p-lee:
* Add test for handler.
* Support logout success URL configuration.

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CLogoutSuccessHandler.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CLogoutSuccessHandler.java
@@ -1,0 +1,62 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See LICENSE in the project root for
+ * license information.
+ */
+package com.microsoft.azure.spring.autoconfigure.b2c;
+
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.web.DefaultRedirectStrategy;
+import org.springframework.security.web.RedirectStrategy;
+import org.springframework.security.web.authentication.logout.LogoutSuccessHandler;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+@Slf4j
+public class AADB2CLogoutSuccessHandler implements LogoutSuccessHandler {
+
+    private String logoutSuccessURL;
+
+    private final AADB2CProperties b2cProperties;
+
+    private final RedirectStrategy redirectStrategy = new DefaultRedirectStrategy();
+
+    public AADB2CLogoutSuccessHandler(@NonNull AADB2CProperties b2cProperties) {
+        this.b2cProperties = b2cProperties;
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CLogoutSuccessHandler.java
```diff
@@ -0,0 +1,62 @@
...
+        this.b2cProperties = b2cProperties;
+    }
+
+    /**
+     * Get the logout success URL.
+     *
+     * @return the URL of logout success.
+     */
+    public String getLogoutSuccessURL() {
+        if (logoutSuccessURL != null) {
```
xscript: `StringUtils.hasText()`?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CLogoutSuccessHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+        return this;
+    }
+
+    @Override
+    public void onLogoutSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication)
+            throws IOException {
+        final String redirectURL = AADB2CURL.getOpenIdLogoutURL(b2cProperties, getLogoutSuccessURL());
+
+        log.info("Http request will be redirect to URL {}.", getLogoutSuccessURL());
```
xscript: Questions:
- Log level should be `info` or `debug`?
- Shall we add the context that logout is successful and HTTP request is redirected to AAD B2C?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CLogoutSuccessHandler.java
Incarnation-p-lee: unnecessary here, `with` method ensure that it will be url format or null.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CLogoutSuccessHandler.java
Incarnation-p-lee: should be debug level, as well as EntryPoint. We can mention that will direct to AAD B2C for expiring the token, but unnecessary for logout successful as it is logoutSuccessHandler.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
```diff
@@ -44,6 +48,14 @@ private static String getEncodedURL(@NonNull String url) {
         }
     }
 
+    public static void validateURL(String url) {
+        try {
+            new URL(url);
+        } catch (MalformedURLException e) {
+            throw new AADB2CConfigurationException("Invalid logoutSuccessUrl: " + url, e);
```
sophiaso: Will this `validateURL` be used to check `logoutSuccessUrl` only? If not, the error message may be misleading.

Incarnation-p-lee: good catch, should update the exception message.

PR #469, opened 2018-11-20
Add aad b2c filter, #464

Incarnation-p-lee:


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CFilter.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public AADB2CFilter(@NonNull AADB2CProperties b2cProperties,
+                        @NonNull AADB2CLogoutSuccessHandler logoutSuccessHandler) {
+        super();
+
+        this.logoutSuccessHandler = logoutSuccessHandler;
+        this.signUpOrInRedirectURL = b2cProperties.getPolicies().getSignUpOrSignIn().getRedirectURI();
+    }
+
+    public String getLogoutSuccessURL() {
```
xscript: Just think of that whether we support relative path? Does Spring Security support relative path as successful logout URL?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CFilterScenario.java
```diff
@@ -0,0 +1,37 @@
...
+    private AADB2CFilterScenarioHandler scenarioHandler;
+
+    AADB2CFilterScenario(@NonNull AADB2CFilterScenarioHandler scenarioHandler) {
+        this.scenarioHandler = scenarioHandler;
+    }
+
+    public static AADB2CFilterScenario resolve(@NonNull HttpServletRequest request, @NonNull AADB2CFilter filter) {
+        final String requestURL = request.getRequestURL().toString();
+
+        if ("GET".equals(request.getMethod()) && requestURL.equals(filter.getSignUpOrInRedirectURL())) {
```
xscript: Is there built-in enum/constant for HTTP method names?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CFilter.java
Incarnation-p-lee: Yes, it support relative path, need to enhance this part for SuccessHandler.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CFilterScenario.java
Incarnation-p-lee: yes, should leverage `org.springframework.http.HttpMethod`.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURL.java
```diff
@@ -48,11 +46,23 @@ private static String getEncodedURL(@NonNull String url) {
...
         try {
             new URL(url);
-        } catch (MalformedURLException e) {
-            throw new AADB2CConfigurationException("Invalid URL: " + url, e);
+            return url;
+        } catch (MalformedURLException ignore) {
+            try {
+                new URI(url);
+                // Absolute URL example: http://localhost:8080/greeting
+                return String.format("%s://%s:%d/%s",
```
xscript: As synced offline, let's make sure there is no gap to build URL by ourselves.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CFilterScenario.java
```diff
@@ -0,0 +1,39 @@
...
+
+    AADB2CFilterScenario(@NonNull AADB2CFilterScenarioHandler scenarioHandler) {
+        this.scenarioHandler = scenarioHandler;
+    }
+
+    public static AADB2CFilterScenario resolve(@NonNull HttpServletRequest request, @NonNull AADB2CFilter filter) {
+        final String requestURL = request.getRequestURL().toString();
+        final String method = request.getMethod();
+
+        if (HttpMethod.GET.matches(method) && requestURL.equals(filter.getSignUpOrInRedirectURL())) {
```
xscript: Ignore case equals?

Incarnation-p-lee: No need here, as Method names are case-sensitive, `get` is another method like user defined.

xscript: What about `requestURL`?

Incarnation-p-lee: In fact, only domain name is case insenstive, so we cannot do that here.
For example, 
https://en.wikipedia.org/wiki/Case_sensitivity
https://en.wikipedia.org/wiki/CASE_SENSITIVITY
Are totally different pages.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CFilterLogoutSuccessHandler.java
```diff
@@ -0,0 +1,20 @@
...
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@NoArgsConstructor
+public class AADB2CFilterLogoutSuccessHandler implements AADB2CFilterScenarioHandler {
+
+    @Override
+    public void handle(HttpServletRequest request, HttpServletResponse response) {
+        throw new UnsupportedOperationException("AADB2CFilterLogoutSuccessHandler implemented yet.");
```
sophiaso: Error message incorrect? `implemented` -> `not implemented`?

Incarnation-p-lee: yes, will remove that in next PR.

PR #470, opened 2018-11-21
initialize validator with configured ResourceRetriever

sophiaso:
## Summary
Token validator is created before ResourceRetriever initialized, which does not take the changed timeout configuration into validator.


sidyes:
@sophiaso Why did you close this?
I am still facing this JWT issue...

PR #490, opened 2018-12-09
allow toggle in Http or Https

mkdirJava:
## Summary
Allow Http / Https toggling 

## Issue Type
- New Feature

## Starter Names
  - storage spring boot starter

## Additional Information
Very minor  change

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/storage/StorageAutoConfiguration.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/storage/StorageAutoConfiguration.java
@@ -31,6 +31,7 @@ import java.util.HashMap;
 public class StorageAutoConfiguration {
     private static final Logger LOG = LoggerFactory.getLogger(StorageAutoConfiguration.class);
     private static final String BLOB_URL = "http://%s.blob.core.windows.net";
+    private static final String BLOB_Https_URL = "https://%s.blob.core.windows.net";
     private static final String USER_AGENT_PREFIX = "spring-storage/";
 
     private final StorageProperties properties;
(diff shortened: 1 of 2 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/storage/StorageAutoConfiguration.java
```diff
@@ -31,6 +31,7 @@
 public class StorageAutoConfiguration {
     private static final Logger LOG = LoggerFactory.getLogger(StorageAutoConfiguration.class);
     private static final String BLOB_URL = "http://%s.blob.core.windows.net";
+    private static final String BLOB_Https_URL = "https://%s.blob.core.windows.net";
```
sophiaso: Use uppercase same as other `static final` constants? i.e., `BLOB_Https_URL` -> `BLOB_HTTPS_URL`

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/storage/StorageAutoConfiguration.java
```diff
@@ -53,7 +54,7 @@ public ServiceURL createServiceUrl(@Autowired(required = false) PipelineOptions
 
         final SharedKeyCredentials credentials = new SharedKeyCredentials(properties.getAccountName(),
                 properties.getAccountKey());
-        final URL blobUrl = new URL(String.format(BLOB_URL, properties.getAccountName()));
+        final URL blobUrl = properties.isEnableHttps() ? new URL(String.format(BLOB_Https_URL, properties.getAccountName())) : new URL(String.format(BLOB_URL, properties.getAccountName()));
```
sophiaso: This line is too long, please restrict the character length within 120 characters.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/storage/StorageProperties.java
```diff
@@ -32,4 +32,8 @@
     @Getter
     @Setter
     private boolean allowTelemetry = true;
+    
+    @Getter
+    @Setter
+    private boolean enableHttps = true;
```
sophiaso: This will change the default behavior compared with previous behavior, i.e., previously, `http` is used, but after this change, `https` will be used. How about set this `enableHttps=false` by default? To avoid affecting other users.

mkdirJava:
Will open another with the alterations

PR #446, opened 2018-10-26
Fixes #436.  Add MediaService configurable connect/read timeout prope…

denniskline:
…rties

## Summary
Attempt to address issue #436  Add MediaService configurable connect/read timeout properties

## Issue Type
- New Feature

## Starter Names
  - media services spring boot starter

## Additional Information
Please let me know if this change is more involved and I misread the issue.

PR #498, opened 2018-12-13
Update version to 2.0.8, #446.

Incarnation-p-lee:


LennartHaggkvistHM:
It's not currently visible on maven.org - is it a delay or do you need to upload the artifacts?

Incarnation-p-lee:
@LennartHaggkvistHM 
Yes, our push to maven vm has return 404 when start job, will try it later. Hope it will be availale soon.

PR #506, opened 2018-12-17
Aad session stateless profile edit and password reset, #464.

Incarnation-p-lee:


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAutoConfiguration.java
```diff
@@ -29,22 +28,69 @@
 @EnableConfigurationProperties(AADB2CProperties.class)
 public class AADB2CAutoConfiguration {
 
-    private final AADB2CProperties b2cProperties;
+    private final AADB2CProperties properties;
```
warrenzhu25: Why rename this here?

Incarnation-p-lee: As in b2c module, the prefix b2c may not provide additional information.

PR #510, opened 2018-12-18
Allow configuration of UserGroup mapping.

DarrenForsythe:
## Summary
Currently the Windows vs Microsoft return a different response. There
is no method of easily the configuring the AAD Starter to use
different keys within the JSON Node.

This PR adds a new nested configuration property to handle this.

This also loosens the validation of the properties for the deprecated
allowed-groups property which could be removed at a later date.

## Issue Type
- New Feature

## Starter Names
## Additional Information


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
```diff
@@ -33,11 +42,12 @@
      * Must be configured when OAuth2 authentication is done in front end
      */
     private String clientSecret;
+
     /**
      * Azure AD groups.
      */
-    @NotEmpty
```
DarrenForsythe: This loosens the validation. I was unsure if adding a new custom validator for the class to check either this or the new property was populated after construction.

DarrenForsythe:
Another idea to extract to an interface and allow different impls that clients could write to lessen the amount of configuration required.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
Incarnation-p-lee: I think the validation put here is for ensuring at least one ad group should be available at runtime. The property without this may makes the configuration from required to optional here.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -92,8 +92,10 @@ private static String getResponseStringFromConn(HttpURLConnection conn) throws I
 
         if (valuesNode != null) {
             valuesNode.forEach(valueNode -> {
-                if (valueNode != null && valueNode.get("objectType").asText().equals("Group")) {
-                    final String objectID = valueNode.get("objectId").asText();
+                if (valueNode != null && valueNode.get(aadAuthenticationProperties.getUserGroup().getKey()).asText()
+                        .equals(aadAuthenticationProperties.getUserGroup().getValue())) {
```
Incarnation-p-lee: suggestion to extract single method for complicated condition, as well as streaming is also recommended here. Like this
`valuesNode.stream().filter(n -> n != null).filter(isGroupEquals).forEach(d -> lUserGroups .add(xxxx))`

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -120,7 +122,9 @@ private static String getResponseStringFromConn(HttpURLConnection conn) throws I
     public Set<GrantedAuthority> convertGroupsToGrantedAuthorities(final List<UserGroup> groups) {
         // Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities
         final Set<GrantedAuthority> mappedAuthorities = groups.stream()
-                .filter(group -> aadTargetGroups.contains(group.getDisplayName()))
+                .filter(group -> aadAuthenticationProperties.getUserGroup().getAllowedGroups()
```
Incarnation-p-lee: the same as line 96, extract complicated condition to method.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
DarrenForsythe: On that understanding I can add a custom validator to cover or validation?

DarrenForsythe: @Incarnation-p-lee validator added to check either property has a value.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
DarrenForsythe: Updated, the value node is not a collection only iterable. Thus the stream support call.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
DarrenForsythe: Updated

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADUserGroupsPropertyValidator.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package com.microsoft.azure.spring.autoconfigure.aad;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+/**
+ * Custom validator to validate that One of the two UserGroup properties are not empty
+ */
+public class AADUserGroupsPropertyValidator
```
sophiaso: The validator annotation way is great, but wonder how about move this validation into the `AADAuthenticationProperties` using a `@PostConstruct` validation? Without introducing a new validator which is only used once in the `AADAuthenticationProperties` class.

DarrenForsythe: ahhh yes, I've actually used that in the past for simple validations. Will refactor

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
```diff
@@ -64,75 +74,41 @@
...
+        private List<String> allowedGroups = new ArrayList<>();
 
-    public int getJwtConnectTimeout() {
-        return jwtConnectTimeout;
-    }
+        /**
+         * Key of the JSON Node to get from the Azure AD response object that will be checked to contain the {@code
+         * azure.activedirectory.user-group.value}  to signify that this node is a valid {@code UserGroup}.
+         */
+        private String key = "objectType";
```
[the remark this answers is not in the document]
DarrenForsythe: 👍

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
```diff
@@ -64,75 +74,41 @@
...
-        return jwtReadTimeout;
-    }
-
-    public void setJwtReadTimeout(int jwtReadTimeout) {
-        this.jwtReadTimeout = jwtReadTimeout;
-    }
+        /**
+         * Key of the JSON Node containing the Azure Object ID for the {@code UserGroup}.
+         */
+        private String objectIDKey = "objectId";
```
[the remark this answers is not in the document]
DarrenForsythe: 👍

DarrenForsythe:
Updated after latest comments. Let me know if there's any other issues to resolve.

Issue #383, opened 2018-07-05
HTTP Response Code 403 while running the sample azure-active-directory-spring-boot-sample 


kaykumar:
## Environment
- Spring boot starter:
    - active directory spring boot starter
  
- OS Type: MacOS
- Java version: 
  - 1.8
 
## Summary
1: Go to azure portal -- > 

https://portal.azure.com/#blade/Microsoft_AAD_IAM/ActiveDirectoryMenuBlade/RegisteredApps

2: Add a new application 
3: set  "oauth2AllowImplicitFlow": true in manifest file
4: create a key under settings
5: Add reply url http://localhost:8080

Configure the sample application "sample azure-active-directory-spring-boot-sample " as described in sample doc

## Reproduce steps

1: Go to localhost:8080
it will perform a SSO with Azure portal
2: It will be unable to load the authorities logging this into the logs:

java.io.IOException: Server returned HTTP response code: 403 for URL: https://graph.windows.net/me/memberOf?api-version=1.6
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1876) ~[na:1.8.0_121]
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1474) ~[na:1.8.0_121]
	at sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:254) ~[na:1.8.0_121]
	at com.microsoft.azure.spring.autoconfigure.aad.AzureADGraphClient.getResponseStringFromConn(AzureADGraphClient.java:72) ~[azure-spring-boot-2.0.4-SNAPSHOT.jar:na]
	at com.microsoft.azure.spring.autoconfigure.aad.AzureADGraphClient.getUserMembershipsV1(AzureADGraphClient.java:59) ~[azure-spring-boot-2.0.4-SNAPSHOT.jar:na]
	at com.microsoft.azure.spring.autoconfigure.aad.AzureADGraphClient.loadUserGroups(AzureADGraphClient.java:87) ~[azure-spring-boot-2.0.4-SNAPSHOT.jar:na]
	at com.microsoft.azure.spring.autoconfigure.aad.AzureADGraphClient.getGroups(AzureADGraphClient.java:83) ~[azure-spring-boot-2.0.4-SNAPSHOT.jar:na]
[...]

PR #511, opened 2018-12-18
Feature/enhance configurability of http calls

DarrenForsythe:
## Summary
Currently the HTTP calls made to the graph API are embedded deep within the code and do not offer an acceptable means to override, change, or enhance these calls. 

This PR breakout the HTTP calls into an interface with two implementations. Add additional testing to cover the URLHTTPConnection and RestTemplate calls.

Additionally this removes the need of Powermock from this starter completely.

## Issue Type
- New Feature

## Starter Names
## Additional Information

What drove this PR was I was getting a different combination of 400, 401, and 200 (with the rest template impl). This _might_ help to provide a workaround to #383 in tandem with #510 as I was seeing the same issues even after the suggested fixes but even without the end goal I feel this provides sensible configuration, and testing additions.


DarrenForsythe:
@Incarnation-p-lee I can do that. I'll likely add the enhanced testing before the refactor to configure the HTTP calls. How do you feel about the overall goal of this PR?

PR #522, opened 2018-12-21
Create claims map instead of reference to jwt, resolve #521.

### LINKED ISSUE #521: Userprincipal reference to unmodified map from JWT
B2C UserPrincipal leverage JWT unmodified claims Map, should keep one copy of this map. As some users may need to adjust the values in some cases.

Incarnation-p-lee:

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/UserPrincipal.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/UserPrincipal.java
@@ -15,10 +15,7 @@ import org.apache.commons.lang3.tuple.Pair;
 
 import java.io.IOException;
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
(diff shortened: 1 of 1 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/UserPrincipal.java
```diff
@@ -127,7 +124,7 @@ public UserPrincipal(@NonNull Pair<JWSObject, JWTClaimsSet> jwtToken, String cod
         this.expireDate = jwtClaimsSet.getExpirationTime();
         this.notValidBeforeDate = jwtClaimsSet.getNotBeforeTime();
         this.issueAtDate = jwtClaimsSet.getIssueTime();
-        this.claim = jwtClaimsSet.getClaims();
```
xscript: Do we need a test case?

Incarnation-p-lee: Yes, I will cover the code for UT later in other PRs.

PR #523, opened 2018-12-26
update Spring-metric-starter dependency

dhaval24:
Update Micrometer dependency to v1.1.1

### DIFF
--- a/azure-spring-boot-parent/pom.xml
+++ b/azure-spring-boot-parent/pom.xml
@@ -33,8 +33,8 @@
         <mockito.core.version>2.8.9</mockito.core.version>
         <powermock.junit4.version>1.7.0</powermock.junit4.version>
         <powermock.api.mockito2.version>1.7.0</powermock.api.mockito2.version>
-        <azuremonitor.micrometer.registry.version>1.1.0</azuremonitor.micrometer.registry.version>
-        <micrometer.core.version>1.1.0</micrometer.core.version>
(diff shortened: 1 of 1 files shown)


azure-spring-boot-parent/pom.xml
```diff
@@ -33,8 +33,8 @@
         <mockito.core.version>2.8.9</mockito.core.version>
         <powermock.junit4.version>1.7.0</powermock.junit4.version>
         <powermock.api.mockito2.version>1.7.0</powermock.api.mockito2.version>
-        <azuremonitor.micrometer.registry.version>1.1.0</azuremonitor.micrometer.registry.version>
-        <micrometer.core.version>1.1.0</micrometer.core.version>
```
Incarnation-p-lee: as we stated before, could you please add this two dependency to https://github.com/VSChina/azure-dependencies-bom/blob/master/pom.xml. It will take over all azure related sdk for version management.

dhaval24: @Incarnation-p-lee I am not sure about it as I have little idea on how azure-dependencies-bom work. Would just by adding these dependencies there and removing it from here, project will compile?

Incarnation-p-lee: yes, just for the version. Simply the dependency bom defines all the version of azure sdk we referenced. As you may know, when we leverage sdk from azure with this dependency management pom. There is no need to take care of the version any more.

Incarnation-p-lee: Yes, I see. Thank you for your effort. We will update that in next release.

PR #552, opened 2019-01-30
Add b2c autoconfiguration for OIDC.

Incarnation-p-lee:


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAutoConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+        private final AADB2CProperties properties;
+
+        public AADB2COidcAutoConfiguration(@NonNull AADB2CProperties properties) {
+            this.properties = properties;
+        }
+
+        @Bean
+        @ConditionalOnMissingBean
+        public ClientRegistrationRepository clientRegistrationRepository() {
+            return new InMemoryClientRegistrationRepository(this.b2cOIDCClientRegistration());
```
sophiaso: Does that mean there will always be only one ClientRegistration available for the user? i.e., B2C here. If it is, do we need to consider support multiple client registrations? e.g., user configured other third party OAuth2 providers.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthorizationRequestResolver.java
```diff
@@ -0,0 +1,58 @@
...
+public class AADB2CAuthorizationRequestResolver implements OAuth2AuthorizationRequestResolver {
+
+    private final AADB2CProperties properties;
+
+    private final OAuth2AuthorizationRequestResolver defaultResolver;
+
+    public AADB2CAuthorizationRequestResolver(ClientRegistrationRepository registrationRepository,
+                                              AADB2CProperties properties) {
+        this.defaultResolver = new DefaultOAuth2AuthorizationRequestResolver(registrationRepository,
+                "/oauth2/authorization");
```
sophiaso: Do we need constant variable for `/oauth2/authorization` here?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAutoConfiguration.java
Incarnation-p-lee: good catch, we should add condititonalon b2c properties here.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthorizationRequestResolver.java
Incarnation-p-lee: I am not sure why this is required, will investigate it and add const then. As you can see, I marked TODO label here.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAutoConfiguration.java
Incarnation-p-lee: Sorry, this bean is under conditionProperty already. See line 62. If user would like to leverage other providers, just remove the property `azure.activedirectory.b2c`.

Incarnation-p-lee: For multi provider support, we may file one issue for this. And will cover that later.

PR #557, opened 2019-02-03
Add password reset and profile edit policy

Incarnation-p-lee:


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CAuthorizationRequestResolver.java
```diff
@@ -17,42 +22,67 @@
...
     @Override
-    public OAuth2AuthorizationRequest resolve(HttpServletRequest request) {
-        Assert.notNull(request, "request cannot be null.");
+    public OAuth2AuthorizationRequest resolve(@NonNull HttpServletRequest request) {
+        if (requestMatcher.matches(request)) {
+            return getB2CAuthorizationRequest(defaultResolver.resolve(request), getRegistrationId(request));
+        }
 
-        return getAADB2CAuthorizationRequest(defaultResolver.resolve(request));
+        return null;
```
warrenzhu25: Returning null is not a good practice

Incarnation-p-lee: yes, but I have to follow the implicit rule with spring security.

Incarnation-p-lee: Will add comments for this.

PR #581, opened 2019-02-16
Add b2c Oidc login Configurer.

Incarnation-p-lee:

### DIFF
--- a/azure-spring-boot-samples/azure-active-directory-b2c-oidc-spring-boot-sample/src/main/java/sample/aad/security/WebSecurityConfiguration.java
+++ b/azure-spring-boot-samples/azure-active-directory-b2c-oidc-spring-boot-sample/src/main/java/sample/aad/security/WebSecurityConfiguration.java
@@ -5,9 +5,7 @@
  */
 package sample.aad.security;
 
-import com.microsoft.azure.spring.autoconfigure.b2c.AADB2CAuthorizationRequestResolver;
-import com.microsoft.azure.spring.autoconfigure.b2c.AADB2CLogoutSuccessHandler;
-import com.microsoft.azure.spring.autoconfigure.b2c.AADB2CProperties;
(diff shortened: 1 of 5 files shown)


azure-spring-boot-samples/azure-active-directory-b2c-oidc-spring-boot-sample/src/main/java/sample/aad/security/WebSecurityConfiguration.java
```diff
@@ -5,43 +5,28 @@
...
-                .logout().logoutSuccessHandler(logoutSuccessHandler)
+                .anyRequest()
+                .authenticated()
                 .and()
-                .oauth2Login()
-                // Customize oauth2 processing url from b2c configuration.
-                .loginProcessingUrl(properties.getLoginProcessingUrl())
-                // Customize oauth2 authorization request.
-                .authorizationEndpoint().authorizationRequestResolver(requestResolver);
+                .apply(configurer)
```
xscript: Question: should we enable it by default when user reference our package as stated in the [doc](https://docs.spring.io/spring-security/site/docs/5.1.4.RELEASE/reference/htmlsingle/#jc-custom-dsls)?
Then user only needs to add dependency and configure `application.yml` without explicit code changes.

Incarnation-p-lee: In fact, I tried that but it looks not a better choose I think, with following reasons:
* The HttpConfigurer almost focus on configuration from `configure(HttpSecurity http)`, it may not works well from Properties configuration like `@EnablePropertiesConfiguration`.
* It may be some annoying when we hidden too much for developers. For convention, developers leverage spring security may write configuration code like
```java
http.authorizeRequests().anyRequest().authenticated().oauth2Login();
```
And we'd better to follow the convention like
```java
http.authorizeRequests().anyRequest().authenticated().apply(aadB2CConfigurer);
```

PR #584, opened 2019-02-18
Add ut for b2c

Incarnation-p-lee:


azure-spring-boot/src/test/java/com/microsoft/azure/spring/autoconfigure/b2c/AADB2CURLTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Java6Assertions.assertThat;
+
+public class AADB2CURLTest {
+
+    @Test
+    public void testGetAuthorizationUrl() {
+        final String expect = "https://fake-tenant.b2clogin.com/fake-tenant.onmicrosoft.com/oauth2/v2.0/authorize";
```
sophiaso: Do we need to add some code about how this url is generated from test constants? As if the test case is failed, others may not understand why it is failed and how to fix it.

Incarnation-p-lee: make sense. But the code for generate the test constant may be the same as implement code. How about add some comments here and link to implement code ?

PR #588, opened 2019-02-19
Add service bus hashed namespace for BI.

Incarnation-p-lee:


azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/servicebus/ServiceBusAutoConfiguration.java
```diff
@@ -66,13 +69,24 @@ public SubscriptionClient subscriptionClient() throws ServiceBusException, Inter
                 properties.getSubscriptionReceiveMode());
     }
 
+    private String getHashNamespace() {
+        final String namespace = properties.getConnectionString()
+                .replaceFirst("^.*//", "") // emit head 'Endpoint=sb://'
+                .replaceAll("\\..*$", ""); // emit tail '${namespace}.xxx.xxx'
+
+        // Namespace can only be letter, number and hyphen, start with letter, end with letter or number,
+        // with length of 6-50.
```
xscript: It is possible that the naming constraint will be extended? Should we stop user from here?

Incarnation-p-lee: make sense, better to just warning here.

PR #609, opened 2019-03-06
Remove dependency of ai sdk.

Incarnation-p-lee:
* Only cosmos, gremlin, media and metrics depends ai sdk.
* Cosmos, gremlin can be removed with new version.


azure-spring-boot/src/main/java/com/microsoft/azure/telemetry/TelemetrySender.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final int RETRY_LIMIT = 3; // Align the retry times with sdk
+
+    private ResponseEntity<String> executeRequest(final TelemetryEventData eventData) {
+        final HttpHeaders headers = new HttpHeaders();
+        final ObjectMapper mapper = new ObjectMapper();
+
+        headers.add(HttpHeaders.CONTENT_TYPE, APPLICATION_JSON.toString());
+
+        try {
+            final RestTemplate restTemplate = new RestTemplate();
```
xscript: Do we need to create a `RestTemplate` every time? `RestTemplate` is thread safe, right?

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelper.java
```diff
@@ -139,14 +137,14 @@ private boolean allowTelemetry(final ConfigurableEnvironment env) {
         return env.getProperty(Constants.AZURE_KEYVAULT_ALLOW_TELEMETRY, Boolean.class, true);
     }
 
-    @PostConstruct
-    private void trackCustomEvent() {
+    private void sendTelemetry() {
```
xscript: Look like it can be extracted as a util method?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilterAutoConfiguration.java
```diff
@@ -72,13 +68,14 @@ public ResourceRetriever getJWTResourceRetriever() {
     }
 
     @PostConstruct
-    private void trackCustomEvent() {
+    private void sendTelemetry() {
         if (aadAuthProps.isAllowTelemetry()) {
             final Map<String, String> events = new HashMap<>();
+            final TelemetrySender sender = new TelemetrySender();
```
xscript: Curious whether we need to create a new instance every time.

azure-spring-boot/src/main/java/com/microsoft/azure/telemetry/TelemetrySender.java
Incarnation-p-lee: make sense, will move to class level.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelper.java
Incarnation-p-lee: In fact it is trade-off here. The `properties` for each starter is quit different, like b2c has `tenant`, storage has `storageAccount`. If extracts util method, only `sender` and `send` can be put into that method, you still to create the `properties` outside and call that util method.
From my person option, that may not make code less or clean.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilterAutoConfiguration.java
Incarnation-p-lee: I decide to create new instance here with following reasons:

* azure-spring-boot has no option for disbale all telemetry, need to keep singleton bean for sender.
* sender is tiny with minor cost when creation.
* according to singleton bean, this way is thread safe.
* the method will be called only once when starter boot up.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/servicebus/ServiceBusAutoConfiguration.java
```diff
@@ -83,12 +82,16 @@ private String getHashNamespace() {
     }
 
     @PostConstruct
-    private void trackCustomEvent() {
-        final HashMap<String, String> events = new HashMap<>();
+    private void sendTelemetry() {
```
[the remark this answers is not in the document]
Incarnation-p-lee: ignore this issue as the method is `@PostConstruct` annotated.

azure-spring-boot-starters/azure-mediaservices-spring-boot-starter/pom.xml
```diff
@@ -30,5 +30,9 @@
             <groupId>com.microsoft.azure</groupId>
             <artifactId>azure-media</artifactId>
         </dependency>
+        <dependency>
```
xscript: why we want AI SDK here?

Incarnation-p-lee: because azure-media depends on ai sdk but our dependency marked ai as optional. Here to make sure ai sdk exist for azure-media, or may meet ClassNotFound from media starter.
Will double check it actually needed.

Incarnation-p-lee: seems no need, will remove this.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelper.java
xscript: I see.

PR #627, opened 2019-03-18
Remove optional dependency ai.

Incarnation-p-lee:
* As azure monitor covers already.


xscript:
How about [here](https://github.com/Microsoft/azure-spring-boot/blob/master/azure-spring-boot-starters/azure-spring-boot-starter/pom.xml#L39-L49)?

Incarnation-p-lee:
Got it, will update ASAP.

PR #677, opened 2019-06-07
Fix properties for specific tenant-id endpoints

arvindn05:
## Summary
The given properties for customizing endpoints for tenant-id do not work with spring boot 2.1.1 or later

## Issue Type
- Bug fixing

## Starter Names

## Additional Information


sophiaso:
@arvindn05 Thanks for opening PR, our latest release used spring boot version `2.1.0.RELEASE`, does this issue only happens for spring boot 2.1.1 or later? Could you also add some reference about why the properties should be changed in this way, e.g., spring boot document or code changes etc.

arvindn05:
i dont think the version matters. The provider name for azure is "azure" and not "azureAuthProvider", so a reference to the configuration inside the hashmap "provider " should be "azure"

see
https://github.com/spring-projects/spring-boot/blob/2.0.x/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientProperties.java

ademeglio:
@arvindn05 , Thanks for finding this. I was wondering why the app continued to redirect to the /common instead of {tenant}. Once I looked at you changes, it made perfect sense. Not sure why this hasn't been merged yet.

Issue #679, opened 2019-06-10
AAD graph API upgrading.


Incarnation-p-lee:
As there are some issues about AAD graph API upgrading, and I summarized some features and issues as below:

* [ ] Support new Graph API.
* [ ] Allow older user leverage old graph API (MS graph) with one EOL time.
* [ ] https://github.com/microsoft/azure-spring-boot/issues/675 
* [ ] https://github.com/microsoft/azure-spring-boot/issues/663
* [ ] https://github.com/microsoft/azure-spring-boot/issues/659
* [ ] https://github.com/microsoft/azure-spring-boot/issues/642

neuqlz:
the v2 feature has merged into the master now.

PR #680, opened 2019-06-11
Update to new graph api, #679

wandrille-hubert:
Update to new graph api

## Summary
Update to the new graph api.  Carry api through security principal (both in oidcUserService flow and AADAuthenticationFilter flow).

## Issue Type
- Update Feature (old graph to new graph)

## Starter Names

## Additional Information


Incarnation-p-lee:
Ref this PR to #679

sophiaso:
@wandrille-hubert Thanks for the contribution, some test case failed, could you check?

wandrille-hubert:
@sophiaso fixed those failing tests.  by the way did you want me to update the endpoints.cn.aadMembershipRestUri and cn.aadGraphApiUri to new graph api (microsoftgraph)?

Incarnation-p-lee:
@wandrille-hubert
Thanks a lot for your contribution. It looks your PR is kind of `Breaking Change`, right ? I mean if I would like to leverage old graph API, it will works well. Somehow one day the old graph will be expired, but it is better to keep both works well, as well as some EOL for old one.

How do u think of it ?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
```diff
@@ -79,9 +79,10 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
                     request.getSession().setAttribute(CURRENT_USER_PRINCIPAL_GRAPHAPI_TOKEN, graphApiToken);
                 }
 
-                final Authentication authentication = new PreAuthenticatedAuthenticationToken(
+                final PreAuthenticatedAuthenticationToken preauth = new PreAuthenticatedAuthenticationToken(
                         principal, null, client.convertGroupsToGrantedAuthorities(principal.getUserGroups()));
-
+                preauth.setDetails(graphApiToken);
+                final Authentication authentication = preauth;
```
Incarnation-p-lee: seems no need to declare authentication.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
```diff
@@ -79,9 +79,10 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
                     request.getSession().setAttribute(CURRENT_USER_PRINCIPAL_GRAPHAPI_TOKEN, graphApiToken);
                 }
 
-                final Authentication authentication = new PreAuthenticatedAuthenticationToken(
+                final PreAuthenticatedAuthenticationToken preauth = new PreAuthenticatedAuthenticationToken(
```
Incarnation-p-lee: `preauth` -> `preAuth`

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -55,9 +55,10 @@ private String getUserMembershipsV1(String accessToken) throws IOException {
 
         final HttpURLConnection conn = (HttpURLConnection) url.openConnection();
         // Set the appropriate header fields in the request header.
-        conn.setRequestProperty("api-version", "1.6");
-        conn.setRequestProperty("Authorization", accessToken);
-        conn.setRequestProperty("Accept", "application/json;odata=minimalmetadata");
+        conn.setRequestMethod("GET");
```
Incarnation-p-lee: avoiding hard code here.
```java
        conn.setRequestMethod(HttpMethod.GET.toString());
        conn.setRequestProperty(HttpHeaders.AUTHORIZATION, String.format("%s %s", OAuth2AccessToken.TokenType.BEARER.getValue(), accessToken);
        conn.setRequestProperty(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON_VALUE);
        conn.setRequestProperty(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_FORM_URLENCODED_VALUE);
```

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/GraphOidcUser.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.security.oauth2.core.oidc.OidcUserInfo;
+import org.springframework.security.oauth2.core.oidc.user.OidcUser;
+import java.util.Collection;
+import java.util.Map;
+
+public class GraphOidcUser implements OidcUser {
+    private final OidcUser delegate;
+    private final String graphApiToken;
+
+    public GraphOidcUser(OidcUser delegate, String token) {
```
Incarnation-p-lee: Could you please explain more about why `token` is required for `UserInfo` ?

azure-spring-boot/src/test/java/com/microsoft/azure/spring/autoconfigure/aad/UserPrincipalTest.java
```diff
@@ -57,16 +57,15 @@ public void getAuthoritiesByUserGroups() throws Exception {
         aadAuthProps.getUserGroup().setAllowedGroups(Collections.singletonList("group1"));
         this.graphClientMock = new AzureADGraphClient(credential, aadAuthProps, endpointsProps);
 
-        stubFor(get(urlEqualTo("/memberOf")).withHeader("Accept", equalTo("application/json;odata=minimalmetadata"))
+        stubFor(get(urlEqualTo("/memberOf")).withHeader("Accept", equalTo("application/json"))
```
Incarnation-p-lee: also avoiding hard code.

azure-spring-boot/src/test/java/com/microsoft/azure/spring/autoconfigure/aad/UserPrincipalTest.java
```diff
@@ -57,16 +57,15 @@ public void getAuthoritiesByUserGroups() throws Exception {
...
                 .willReturn(aResponse().withStatus(200).withHeader("Content-Type", "application/json")
                         .withBody(Constants.USERGROUPS_JSON)));
 
         assertThat(graphClientMock.getGrantedAuthorities(Constants.BEARER_TOKEN)).isNotEmpty()
                 .extracting(GrantedAuthority::getAuthority).containsExactly("ROLE_group1");
 
-        verify(getRequestedFor(urlMatching("/memberOf")).withHeader("Authorization", equalTo(accessToken))
-                .withHeader("Accept", equalTo("application/json;odata=minimalmetadata"))
-                .withHeader("api-version", equalTo("1.6")));
+        verify(getRequestedFor(urlMatching("/memberOf")).withHeader("Authorization", equalTo("Bearer " + accessToken))
```
Incarnation-p-lee: the same as above.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/GraphOidcUser.java
wandrille-hubert: This was requested in order to be able to get/utilize the api token generated by calling/utilizing the spring boot security principal.

Incarnation-p-lee: from my understanding, `userinfo` should not contain any `token` like information. If you would like to get `token`, you should obtain that from SecurityContextHolder.

wandrille-hubert:
@Incarnation-p-lee yes I agree that this is a breaking change as it is swapping the old api for the new.  Would we want to set a flag in the settings to let users choose which api to go with/or would we want to grab both api tokens?  seems like only getting one api token might be more favorable.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/GraphOidcUser.java
wandrille-hubert: so I assume the token to get from the SecurityContextHolder is as such:
`Authentication authentication = SecurityContextHolder.getContext().getAuthentication(); DefaultOidcUser principal = (DefaultOidcUser) authentication.getPrincipal(); 
OidcIdToken idToken = principal.getIdToken(); 
String accessToken = idToken.getTokenValue(); `

but I notice that this token gotten does not equal the api token, and also when trying to use this token against the graph.microsoft.com, get an access token validation failure.
might I be missing something/or does additional code need to be added in to set the graph api token in the SecurityContextHolder?

Incarnation-p-lee:
@wandrille-hubert
I may prefer somewhat like flag to reach this, we can make it to EOL once old graph expires.
@sophiaso How do u think of it ?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/GraphOidcUser.java
Incarnation-p-lee: I think the point is that if we should put `token` here, instead of the `token` validation failure, so we add the `token` to user info. Unless there is only way we can do, I may need to try your code later.

sophiaso:
@wandrille-hubert Over the two options, prefer to ` set a flag in the settings to let users choose which api to go with`, thanks.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADOAuth2UserService.java
```diff
@@ -74,8 +74,8 @@ public OidcUser loadUser(OidcUserRequest userRequest) throws OAuth2Authenticatio
 
         // Create a copy of oidcUser but use the mappedAuthorities instead
         oidcUser = new DefaultOidcUser(mappedAuthorities, oidcUser.getIdToken(), getUserNameAttrName(userRequest));
-        final GraphOidcUser graphOidcUser = new GraphOidcUser(oidcUser, graphApiToken);
-
+        final GraphOidcUser graphOidcUser = new GraphOidcUser(oidcUser, graphApiToken);
```
Incarnation-p-lee: looks some space after `;`.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADOAuth2UserService.java
```diff
@@ -74,8 +74,8 @@ public OidcUser loadUser(OidcUserRequest userRequest) throws OAuth2Authenticatio
 
         // Create a copy of oidcUser but use the mappedAuthorities instead
         oidcUser = new DefaultOidcUser(mappedAuthorities, oidcUser.getIdToken(), getUserNameAttrName(userRequest));
-        final GraphOidcUser graphOidcUser = new GraphOidcUser(oidcUser, graphApiToken);
-
+        final GraphOidcUser graphOidcUser = new GraphOidcUser(oidcUser, graphApiToken);   
+
```
Incarnation-p-lee: new line with all space.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/ServiceEndpoints.java
```diff
@@ -14,6 +14,9 @@
 public class ServiceEndpoints {
     private String aadSigninUri;
     private String aadGraphApiUri;
+    private String aadGraphApiUriNew;
```
Incarnation-p-lee: the word `new` is unclear, could you please add more information about this ?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/ServiceEndpoints.java
```diff
@@ -14,6 +14,9 @@
...
     private String aadKeyDiscoveryUri;
     private String aadMembershipRestUri;
+    private String aadMembershipRestUriNew;
+    private String aadUseNewGraphApi;
```
Incarnation-p-lee: you can use `bool` directly, and the name `aadUseNewGraphApi` looks like action not true of false.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -177,12 +207,19 @@ public AuthenticationResult acquireTokenForGraphApi(String idToken, String tenan
         AuthenticationResult result = null;
         ExecutorService service = null;
         try {
+            final String graphApiUriString;
```
Incarnation-p-lee: you can extract line 211 ~ 216 to class `serviceEndpoints`.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -55,15 +55,28 @@ public AzureADGraphClient(ClientCredential clientCredential, AADAuthenticationPr
...
-        conn.setRequestProperty(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_FORM_URLENCODED_VALUE);
+
+        if (useNewGraphApi) {
+            conn.setRequestMethod(HttpMethod.GET.toString());
+            conn.setRequestProperty(HttpHeaders.AUTHORIZATION, 
+                                    String.format("%s %s", OAuth2AccessToken.TokenType.BEARER.getValue(), accessToken));
+            conn.setRequestProperty(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON_VALUE);
+            conn.setRequestProperty(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_FORM_URLENCODED_VALUE);
+        } else {
+            conn.setRequestProperty("api-version", "1.6");
```
Incarnation-p-lee: No RequestMethod ?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
```diff
@@ -31,6 +31,11 @@
     private UserGroupProperties userGroup = new UserGroupProperties();
 
 
+    /**
+     * Default UserGroup configuration for old graph api.
+     */
+    private UserGroupPropertiesOld userGroupOld = new UserGroupPropertiesOld();
```
sophiaso: I'm more preferred we can have only one `userGroup` field, instead of using `UserGroupProperties` and `UserGroupPropertiesOld`, how about we have an abstract `AbstractUserGroupProperties`(naming can change), and make an AAD graph user group properties and another microsoft graph user group properties which both extend the `AbstractUserGroupProperties`.

azure-spring-boot/src/main/resources/serviceEndpoints.properties
```diff
@@ -4,6 +4,8 @@ azure.service.endpoints.cn.aadKeyDiscoveryUri=https://login.partner.microsoftonl
 azure.service.endpoints.cn.aadMembershipRestUri=https://graph.chinacloudapi.cn/me/memberOf?api-version=1.6
 azure.service.endpoints.global.aadSigninUri=https://login.microsoftonline.com/
 azure.service.endpoints.global.aadGraphApiUri=https://graph.windows.net/
+azure.service.endpoints.global.aadGraphApiUriNew=https://graph.microsoft.com/
```
sophiaso: Is it possible we have two `serviceEndpoints.properties` file? e.g., `aadGraphServiceEndpoints.properties` and `microsoftGraphServiceEndpoints.properties`? Each of them have the same key list, instead of mixing them in one file?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
wandrille-hubert: so I have gone ahead and changed it to an abstract class with two different property classes that extend this abstract class.  the one issue I was running into was if there are override properties, then I was not able to get these applied.  i tried a multitude of various annotations/etc but couldnt get anything to stick.  the only way i could get this to work was by initializing the usergroup to one of those two property classes at the beginning, and then if setting the usergroup to the other property in the case that microsoft graph api is to be used, then to check to see if any properties were overridden, and if so, to path this into that property class initializer.  all this can be seen in the AADAuthenticationProperties.java file.

azure-spring-boot/src/main/resources/serviceEndpoints.properties
wandrille-hubert: so i went ahead and kept the serviceEndpoints.properties for the aad graph, and then microsoftGraphServiceEndpoints.properties for the microsoft graph api.  in this file, i set the environment as global-microsoftgraph.  So a property is of the form: azure.service.endpoints.global-microsoftgraph.aadMembershipRestUri.

let me know if this is acceptable or you had envisioned a different approach.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -93,15 +107,16 @@ private static String getResponseStringFromConn(HttpURLConnection conn) throws I
         final JsonNode valuesNode = rootNode.get("value");
 
         if (valuesNode != null) {
-
+            //final boolean microsoftGraphApiBool = serviceEndpoints.isAadMicrosoftGraphApiBool();
```
saragluna: remove commented out code

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -93,15 +107,16 @@ private static String getResponseStringFromConn(HttpURLConnection conn) throws I
...
             lUserGroups
-                    .addAll(StreamSupport.stream(valuesNode.spliterator(), false).filter(this::isMatchingUserGroupKey)
-                            .map(node -> {
-                                final String objectID = node.
-                                        get(aadAuthenticationProperties.getUserGroup().getObjectIDKey()).asText();
-                                final String displayName = node.get("displayName").asText();
-                                return new UserGroup(objectID, displayName);
-                            }).collect(Collectors.toList()));
+            .addAll(StreamSupport.stream(valuesNode.spliterator(), false).filter(this::isMatchingUserGroupKey)
+                    .map(node -> {
```
saragluna: seems only changed indentation

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -116,8 +131,9 @@ private static String getResponseStringFromConn(HttpURLConnection conn) throws I
      * @return true if the json node contains the correct key, and expected value to identify a user group.
      */
     private boolean isMatchingUserGroupKey(final JsonNode node) {
+        //final boolean microsoftGraphApiBool = serviceEndpoints.isAadMicrosoftGraphApiBool();
```
saragluna: remove commented out codes

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -116,8 +131,9 @@ private static String getResponseStringFromConn(HttpURLConnection conn) throws I
...
         return node.get(aadAuthenticationProperties.getUserGroup().getKey()).asText()
-                .equals(aadAuthenticationProperties.getUserGroup().getValue());
```
saragluna: only changed indentation

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -158,7 +174,9 @@ private boolean isMatchingUserGroupKey(final JsonNode node) {
      * @return true if either of the allowed-groups or active-directory-groups contains the UserGroup display name
      */
     private boolean isValidUserGroupToGrantAuthority(final UserGroup group) {
-        return aadAuthenticationProperties.getUserGroup().getAllowedGroups().contains(group.getDisplayName())
+        //final boolean microsoftGraphApiBool = serviceEndpoints.isAadMicrosoftGraphApiBool();
```
saragluna: remove commented out code

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -158,7 +174,9 @@ private boolean isMatchingUserGroupKey(final JsonNode node) {
...
+        return aadAuthenticationProperties.getUserGroup().getAllowedGroups()
```
saragluna: only changed indentation

saragluna:
@wandrille-hubert It seems like the Codacy quality review failed. Since new graph API and new config introduced, we'd better update the document and sample projects too.

wandrille-hubert:
@saragluna for the sample projects, did we want to update the current ones to use the new graph api, or create new samples that utilize the graph api and keep the current ones as is?

neuqlz:
@wandrille-hubert  Creating new samples that utilize the graph APIs and keeping the current ones are a better choice. And you also need to fix the Codacy Quality Review issues related to your PR.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
```diff
@@ -87,12 +87,12 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
-                final Authentication authentication = new PreAuthenticatedAuthenticationToken(
+                final PreAuthenticatedAuthenticationToken preAuth = new PreAuthenticatedAuthenticationToken(
                         principal, null, client.convertGroupsToGrantedAuthorities(principal.getUserGroups()));
-
-                authentication.setAuthenticated(true);
-                log.info("Request token verification success. {}", authentication);
-                SecurityContextHolder.getContext().setAuthentication(authentication);
+                preAuth.setDetails(graphApiToken);
+                preAuth.setAuthenticated(true);
+                log.info("Request token verification success. {}", preAuth);
```
neuqlz: Why do you need to put graphApi Token in PreAuthenticatedAuthenticationToken Object? From my perspective, you shouldn't put graphToken here. If you need this token to invoke graph APIs, you can invoke com.microsoft.azure.spring.autoconfigure.aad.AzureADGraphClient#acquireTokenForGraphApi this method to get the new token, rather than use the old graphToken cached in the Authentication Object.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADOAuth2UserService.java
```diff
@@ -74,8 +74,9 @@ public OidcUser loadUser(OidcUserRequest userRequest) throws OAuth2Authenticatio
 
         // Create a copy of oidcUser but use the mappedAuthorities instead
         oidcUser = new DefaultOidcUser(mappedAuthorities, oidcUser.getIdToken(), getUserNameAttrName(userRequest));
+        final GraphOidcUser graphOidcUser = new GraphOidcUser(oidcUser, graphApiToken);
```
neuqlz: In this area, there is the same problem with the problem in AADAuthenticationFilter.java this Class, you shouldn't put graphToken in the OidcUser object, if you need to use it, you can get a new one again with the com.microsoft.azure.spring.autoconfigure.aad.AzureADGraphClient#acquireTokenForGraphApi method. What do you think about it?

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
```diff
@@ -81,6 +80,42 @@
      */
     private boolean allowTelemetry = true;
 
+    public void setUserGroupWithBool(boolean microsoftGraphApi) {
+        if (microsoftGraphApi) {
+            if (!MicrosoftGraphUserGroupProperties.class.isInstance(this.userGroup)) {
+                final AADGraphUserGroupProperties tempProperties = new AADGraphUserGroupProperties();
```
neuqlz: To be honest, I am a little confused with this piece of code, When the switch is on(using the version 2 graph APIs) and current variable userGroup is not the instance of MicrosoftGraphUserGroupProperties, why not just assign a new object of MicrosoftGraphUserGroupProperties to the local variable userGroup? Can you explain it to me? Thank you.

wandrille-hubert: @neuqlz 
so the one issue I was running into was if there are override properties (in a property file) in the UserGroup, then I was not able to get these applied. i tried a multitude of various annotations/etc but couldnt get anything to stick. the only way i could get this to work was by initializing the usergroup to one of those two property classes at the beginning, and then if setting the usergroup to the other property in the case that microsoft graph api is to be used, then to check to see if any properties were overridden, and if so, then apply them with such a "weird" function.

ideally, if the override properties would get applied to that property, then wouldn't need to do this.  i even tried separating out the MicrosoftGraphUserGroupProperties to its own file, but still couldnt get the override properties to get applied.  (The test where this comes up is the testUserGroupPropertiesAreOverriden in the AADAuthenticationAutoConfigurationTest.java).

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
wandrille-hubert: so the reason I had put the token in the Authentication Object was to be able to persist the token in the spring security principal object: 
String token = (String)SecurityContextHolder.getContext().getAuthentication().getDetails();

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADOAuth2UserService.java
wandrille-hubert: similarly as above, in order to be able to get the token from the principal: 
GraphOidcUser graphUser = (GraphOidcUser)SecurityContextHolder.getContext().getAuthentication().getPrincipal();

String token = graphUser.getGraphApiToken();

please let me know your thoughts/if this should be removed (similarly for the above one)

wandrille-hubert:
@neuqlz ok will get some new samples created.

also the codacy quality review, will take care of that once we resolve the converstaion on .../src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
as that is the reason for that quality to fail.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationFilter.java
neuqlz: Yeah, I got your reason. But my point is that we shouldn't cache the graphAPIs token here, if you want to use the token, you have many ways to get it, such as getting it from the session, getting it by invoking AzureADGraphClient#acquireTokenForGraphApi. 

Another question: what do you do by using the graphAPIs token? The granted authorities Infos have been put in the context already.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADOAuth2UserService.java
neuqlz: Change the OidcUser's implementation from DefaultOidcUser to GraphOidcUser  is not safe, it may cause some incompatibility problem for the old user who uses like this:
DefaultOidcUser graphUser = (DefaultOidcUser)SecurityContextHolder.getContext().getAuthentication().getPrincipal();
And also the token should not be put in the OidcUser, it's not safe.

If you want to get the graphAPIs token. You have many choices to get it:
method one:
 you can put graphAPIs token in the session, than you can get it from the session just like the way we do in AADAuthenticationFilter.class

method two:
you can get the graphAPIs token by AzureADGraphClient#acquireTokenForGraphApi:
graphApiToken = graphClient.acquireTokenForGraphApi(idToken.getTokenValue(),aadAuthProps.getTenantId()).getAccessToken();

azure-spring-boot/src/main/resources/microsoftGraphServiceEndpoints.properties
```diff
@@ -0,0 +1,9 @@
+azure.service.endpoints.cn-microsoftgraph.aadSigninUri=https://login.partner.microsoftonline.cn/
+azure.service.endpoints.cn-microsoftgraph.aadGraphApiUri=https://microsoftgraph.chinacloudapi.cn/
+azure.service.endpoints.cn-microsoftgraph.aadKeyDiscoveryUri=https://login.partner.microsoftonline.cn/common/discovery/keys
+azure.service.endpoints.cn-microsoftgraph.aadMembershipRestUri=https://microsoftgraph.chinacloudapi.cn/v1.0/me/memberOf
+azure.service.endpoints.global-microsoftgraph.aadSigninUri=https://login.microsoftonline.com/
+azure.service.endpoints.global-microsoftgraph.aadGraphApiUri=https://graph.microsoft.com/
+azure.service.endpoints.global-microsoftgraph.aadKeyDiscoveryUri=https://login.microsoftonline.com/common/discovery/keys/
+azure.service.endpoints.global-microsoftgraph.aadMembershipRestUri=https://graph.microsoft.com/v1.0/me/memberOf
+azure.service.endpoints.global-microsoftgraph.aadMicrosoftGraphApiBool=true
```
neuqlz: just put these endpoints in the file serviceEndpoints.properties. There is no need to separate them into two files.

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAuthenticationProperties.java
neuqlz: For the userGroupProperties, why you need to separate it into two child classes? (AADGraphUserGroupProperties and MicrosoftGraphUserGroupProperties) I think you can just use UserGroupProperties. Your purpose is that when you use v2 graph APIs, the variables in UserGroupProperties should be the v2 version value. So you can just configurate it by application.properties, just like:

azure.activedirectory.user-group.allowed-groups=another_group,third_group
azure.activedirectory.user-group.key=@odata.type
azure.activedirectory.user-group.value=#microsoft.graph.group
azure.activedirectory.user-group.object-id-key=id

wandrille-hubert: ok sounds good, will change it to that

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADOAuth2UserService.java
wandrille-hubert: ok makes sense, have gone ahead and removed GraphOidcUser

azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AzureADGraphClient.java
```diff
@@ -51,13 +54,23 @@ public AzureADGraphClient(ClientCredential clientCredential, AADAuthenticationPr
     }
 
     private String getUserMembershipsV1(String accessToken) throws IOException {
+        final boolean microsoftGraphApiBool = serviceEndpoints.isAadMicrosoftGraphApiBool();
```
neuqlz: I think there is no need to use an extra flag to judge if the current service uses v1 or v2 graphAPIs. Just use the environment. For example:

    private final String V2_VERSION_ENV_FLAG = "microsoftgraph";
    private boolean aadMicrosoftGraphApiBool;
    

    public AzureADGraphClient(ClientCredential clientCredential, AADAuthenticationProperties aadAuthProps,
                              ServiceEndpointsProperties serviceEndpointsProps) {
        this.clientId = clientCredential.getClientId();
        this.clientSecret = clientCredential.getClientSecret();
        this.aadAuthenticationProperties = aadAuthProps;
        this.serviceEndpoints = serviceEndpointsProps.getServiceEndpoints(aadAuthProps.getEnvironment());

        this.initAADMicrosoftGraphApiBool(aadAuthProps.getEnvironment()));
    }

    private void initAADMicrosoftGraphApiBool(String endpointEnv) {
        this.aadMicrosoftGraphApiBool = false;
        if (endpointEnv.contains(V2_VERSION_ENV_FLAG)) {
            this.aadMicrosoftGraphApiBool = true;
        }
    }

azure-spring-boot/src/main/resources/serviceEndpoints.properties
```diff
@@ -2,8 +2,17 @@ azure.service.endpoints.cn.aadSigninUri=https://login.partner.microsoftonline.cn
 azure.service.endpoints.cn.aadGraphApiUri=https://graph.chinacloudapi.cn/
 azure.service.endpoints.cn.aadKeyDiscoveryUri=https://login.partner.microsoftonline.cn/common/discovery/keys
 azure.service.endpoints.cn.aadMembershipRestUri=https://graph.chinacloudapi.cn/me/memberOf?api-version=1.6
+azure.service.endpoints.cn-microsoftgraph.aadSigninUri=https://login.partner.microsoftonline.cn/
+azure.service.endpoints.cn-microsoftgraph.aadGraphApiUri=https://microsoftgraph.chinacloudapi.cn/
+azure.service.endpoints.cn-microsoftgraph.aadKeyDiscoveryUri=https://login.partner.microsoftonline.cn/common/discovery/keys
+azure.service.endpoints.cn-microsoftgraph.aadMembershipRestUri=https://microsoftgraph.chinacloudapi.cn/v1.0/me/memberOf
 azure.service.endpoints.global.aadSigninUri=https://login.microsoftonline.com/
```
neuqlz: I think you should add some comments here to explain xx-microsoftgraph is associated with v2 version or change the properties' keys to more obvious keys, just like: azure.service.endpoints.cn-v2-graph.aadSigninUri

neuqlz:
@wandrille-hubert And I also think there need some guides which teach users how to use the version2 graphAPIs. Could you add a section which guides users how to use the version2 graphAPIs into the doc(azure-spring-boot-starters/azure-active-directory-spring-boot-starter/README.md) ? Thanks a lot for your contribution.

PR #872, opened 2020-04-26
Skip AAD internal filter when is already authenticated or token not issued by AAD

saragluna:
## Summary

When `AADAuthenticationFilter` or `AADAppRoleStatelessAuthenticationFilter` is auto-configured it will take action on every bearer token. This PR will skip the processing when the current context is already authenticated or the token is not issued by Azure AD.

## Issue Type

- New Feature

## Starter Names
  
## Additional Information

### DIFF
--- a/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAppRoleStatelessAuthenticationFilter.java
+++ b/azure-spring-boot/src/main/java/com/microsoft/azure/spring/autoconfigure/aad/AADAppRoleStatelessAuthenticationFilter.java
@@ -52,27 +52,8 @@ public class AADAppRoleStatelessAuthenticationFilter extends OncePerRequestFilte
         final String authHeader = request.getHeader(HttpHeaders.AUTHORIZATION);
         boolean cleanupRequired = false;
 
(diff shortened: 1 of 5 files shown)


nikhilingole:
Hey @saragluna / @jialindai  - Any updates on this, when would it be part of a release. 
Thanks.

saragluna:
@nikhilingole This will be out in our next release, which is two weeks later.

nikhilingole:
Thanks @saragluna for the update and fixing the issue.

As a side comment - Shouldn't the response from the filter be 401/403 if the Token in invalid/expired etc instead of 5xx.

morev-a:
Hi @saragluna, I thing you have to look on AbstractAuthenticationProcessingFilter. 
> AbstractAuthenticationProcessingFilter - a base Filter used for authentication.
https://docs.spring.io/spring-security/site/docs/current/reference/html5/#servlet-authentication-abstractprocessingfilter

The issue here is that AADAppRoleStatelessAuthenticationFilter is extended from OncePerRequestFilter. It's executed on EVERY request. That is absolutely wrong. It ignores configuration extended from WebSecurityConfigurerAdapter. That's why there is no any way to configure multiple authentication providers.

PR #880, opened 2020-05-04
Deliver multiple KeyVault support

mnriem:
## Summary

Delivers support for multiple keyvaults

## Issue Type
- New Feature

## Starter Names

## Additional Information

### DIFF
--- a/azure-spring-boot-starters/azure-keyvault-secrets-spring-boot-starter/README.md
+++ b/azure-spring-boot-starters/azure-keyvault-secrets-spring-boot-starter/README.md
@@ -117,3 +117,8 @@ azure.keyvault.keyvault2.client-id=put-a-azure-client-id-here
 azure.keyvault.keyvault2.client-key=put-a-azure-client-key-here
 azure.keyvault.keyvault2.tenant-id=put-a-azure-tenant-id-here
 ```
+
+Note if you decide to use multiple key vault support and you already have an
+existing configuration, please make sure you migrate that configuration to the
+multiple key vault variant. Mixing multiple key vaults with an existing single
+key vault configuration is a non supported scenario.
--- a/azure-spring-boot-starters/azure-keyvault-secrets-spring-boot-starter/README.md
+++ b/azure-spring-boot-starters/azure-keyvault-secrets-spring-boot-starter/README.md
@@ -96,4 +96,24 @@ azure.keyvault.allow.telemetry=false
 When telemetry is enabled, an HTTP request will be sent to URL `https://dc.services.visualstudio.com/v2/track`. So please make sure it's not blocked by your firewall.    
 Find more information about Azure Service Privacy Statement, please check [Microsoft Online Services Privacy Statement](https://www.microsoft.com/en-us/privacystatement/OnlineServices/Default.aspx). 
 
+## Multiple Key Vault support
 
+If you want to use multiple key vaults you need to define names for each of the
+key vaults you want to use and in which order the key vaults should be consulted.
+If a property exists in multiple key vaults the order determine which value you
+will get back.
+
+The example below shows a setup for 2 key vaults, named `keyvault1` and
+`keyvault2`. The order specifies that `keyvault1` will be consulted first.
+
+```
+azure.keyvault.order=keyvault1,keyvault2
+azure.keyvault.keyvault1.uri=put-a-azure-keyvault-uri-here
+azure.keyvault.keyvault1.client-id=put-a-azure-client-id-here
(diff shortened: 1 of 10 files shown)


azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelper.java
```diff
@@ -46,16 +45,28 @@ public KeyVaultEnvironmentPostProcessorHelper(final ConfigurableEnvironment envi
...
+    /**
+     * Add a key vault property source.
+     *
+     * <p>
+     * The normalizedName is used to target a specific key vault (note if the
+     * name is the empty string it works as before with only one key vault
+     * present).
+     * </p>
+     *
+     * @param postfix the postfix.
```
saragluna, line 57: inaccurate parameter name `postfix`

azure-spring-boot-tests/azure-spring-boot-test-keyvault/pom.xml
```diff
@@ -15,6 +15,41 @@
...
+                                <requireEnvironmentVariable>
+                                    <variableName>AZURE_TENANT</variableName>
+                                    <message>Set the AZURE_TENANT environment variable to your Azure Tenant ID</message>
+                                </requireEnvironmentVariable>
+                                <requireEnvironmentVariable>
+                                    <variableName>AZURE_SUBSCRIPTION</variableName>
+                                    <message>Set the AZURE_SUBSCRIPTION environment variable to your Azure Subscription ID</message>
+                                </requireEnvironmentVariable>
+                                <requireEnvironmentVariable>
+                                    <variableName>AZURE_CLIENT_ID</variableName>
```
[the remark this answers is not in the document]
mnriem: Added it to the POM

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultEnvironmentPostProcessorHelper.java
mnriem: Changed it to `normalizedName` and added a more detailed explanation of what the normalizedName is.

mnriem:
When you say the document which one do you mean? 

On the order yes we need to clarify that the order of the .names attribute determines the order in which the respective KeyVaults are consulted. 

On the naming convention I opted for the key vault name at the beginning as it was easier/cleaner in the implementation to add it, but I have no preference. So if it needs to be changed just say the word.

saragluna:
Sorry for not mentioning where to add them. A section describing the new feature added to the [README.md](https://github.com/microsoft/azure-spring-boot/blob/master/azure-spring-boot-starters/azure-keyvault-secrets-spring-boot-starter/README.md) will be enough. 

If the `<keyvault-name>` has no usage other than distinguishing different configurations, we'd better use array to list these configurations, in which case the order could be expressed by their order showing up in the array. 

If we want to use the `<keyvault-name>` pattern,  I think `azure.keyvault.<keyvault-name>.xxx` is better even though the implementation isn't clean/compact.

mnriem:
OK I will change it to use to the ```azure.keyvault.<keyvault-name>.xxx` syntax. I will rename the `.names` property to `.order` as it expresses what it is used for from the customers perspective, it is required however as it is also used to whether or not to enable multiple key vault support or not.

saragluna:
There're some unit tests failed.

azure-spring-boot-starters/azure-keyvault-secrets-spring-boot-starter/README.md
```diff
@@ -96,4 +96,24 @@ azure.keyvault.allow.telemetry=false
 When telemetry is enabled, an HTTP request will be sent to URL `https://dc.services.visualstudio.com/v2/track`. So please make sure it's not blocked by your firewall.    
 Find more information about Azure Service Privacy Statement, please check [Microsoft Online Services Privacy Statement](https://www.microsoft.com/en-us/privacystatement/OnlineServices/Default.aspx). 
 
+## Multiple Key Vault support
```
saragluna, line 99: What if the user also sets the `azure.keyvault.uri`, what's the order between them?

mnriem: The order would be dependent on the order in the file, but I feel we should caution a customer against doing that. If they opt for multiple key vaults they should only use it that way so the order is guaranteed. I will add this to the documentation.

PR #896, opened 2020-05-21
Keyvault Spring Boot Actuator support

mnriem:
## Summary
Spring Boot Actuator support for Key Vault

## Issue Type
- New Feature

## Starter Names

## Additional Information


azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultPropertySource.java
```diff
@@ -8,9 +8,9 @@
 
 import static com.microsoft.azure.keyvault.spring.Constants.AZURE_KEYVAULT_PROPERTYSOURCE_NAME;
 
-import org.springframework.core.env.PropertySource;
+import org.springframework.core.env.EnumerablePropertySource;
 
-public class KeyVaultPropertySource extends PropertySource<KeyVaultOperation> {
+public class KeyVaultPropertySource extends EnumerablePropertySource<KeyVaultOperation> {
```
saragluna, line 13: May I ask why we change this back to EnumerablePropertySource here? Since @chenrujun changed this in #882.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
```diff
@@ -113,29 +116,51 @@ private boolean needRefreshSecretKeys() {
...
+     */
+    boolean isUp() {
+        boolean up;
+        try {
+            final Response<KeyVaultSecret> response = keyVaultClient
+                    .getSecretWithResponse("should_not_be_empty", null, Context.NONE);
+            up = response.getStatusCode() < 500;
+        } catch (HttpRequestException | ResourceNotFoundException e) {
+            up = true;
+        } catch (RuntimeException re) {
```
saragluna, line 161: Perhaps add some log info here?

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
```diff
@@ -113,29 +116,51 @@ private boolean needRefreshSecretKeys() {
...
+                            secretNameList.add(secretName);
+                        });
+                        return secretNameList;
+                    })
+                    .map(Collection::stream)
+                    .orElseGet(Stream::empty)
+                    .map(this::toKeyVaultSecretName)
+                    .distinct()
+                    .collect(Collectors.toList());
+        } catch (RuntimeException e) {
```
saragluna, line 135: Why try to catch runtime exception here?

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultPropertySource.java
mnriem: For it to show up in the /env actuator endpoint it needs to be an EnumerablePropertySource.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
mnriem: Will do!

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
mnriem: The client throws these when it has an issue connecting to the key vault and then the application just crashes. When the key vault is unavailable (for whatever reason) the application should not crash.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultPropertySource.java
chenrujun: @mnriem Thank you for your work. 

For `EnumerablePropertySource` , We have a case like this: When we have many properties, for example, 150 properties. If we use EnumerablePropertySource, it will cost a lot of time(few minutes, maybe longer) to read all of these properties. And some spring framework logic will get all propertValues more than one time. Here is an example: [spring-config get all propertValues](https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-context/src/main/java/org/springframework/cloud/bootstrap/encrypt/EnvironmentDecryptApplicationInitializer.java#L230-L235)


Could you please cache `propertyVaules` , just like cache `propertyNames`.

1. Use same config item to set the expire time.
2. Set default expire time to 1 minute.

mnriem: I am in the process of addressing performance issues in PR #905, in which indeed the keys and values are cached.

mnriem:
> These three files got deleted accidentally:
> 
> * azure-spring-boot-starters/azure-mediaservices-spring-boot-starter/README.md
> * azure-spring-boot-starters/azure-servicebus-spring-boot-starter/README.md
> * azure-spring-boot-starters/azure-storage-spring-boot-starter/README.md

I did not remove those, please coordinate this with whomever removed them.

chenrujun:
Hi, @mnriem .
Please resolve the conflicts. 🙏

chenrujun:
Hi, @mnriem .

> I did not remove those, please coordinate this with whomever removed them.

Seems you deleted these files in [your commit](https://github.com/microsoft/azure-spring-boot/pull/896/commits/b2373581b04856e7afff2765dc5c6e626ea28012)

PR #904, opened 2020-05-29
upgrade jms underlying library with azure-servicebus-jms-0.0.1

yiliuTo:
Replace the underlying library of jms from org.apache.qpid: qpid-jms-client to com.microsoft.azure: azure-servicebus-jms

### DIFF
--- a/azure-spring-boot-parent/pom.xml
+++ b/azure-spring-boot-parent/pom.xml
@@ -33,8 +33,8 @@
         <mockito.core.version>2.8.9</mockito.core.version>
         <wiremock-standalone.version>2.19.0</wiremock-standalone.version>
         <commons-io.version>2.3</commons-io.version>
-        <qpid-jms-client.version>0.43.0</qpid-jms-client.version>
         <junit-params.version>1.1.1</junit-params.version>
+        <azure-servicebus-jms.version>0.0.1</azure-servicebus-jms.version>
     </properties>
 
     <profiles>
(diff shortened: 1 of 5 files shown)


azure-spring-boot-starters/azure-servicebus-jms-spring-boot-starter/pom.xml
```diff
@@ -34,10 +34,13 @@
         </dependency>
 
         <!--Qpid-->
+<!--        <dependency>-->
```
saragluna, line 37: We should remove this.

azure-spring-boot-bom/pom.xml
```diff
@@ -118,6 +119,11 @@
                 <artifactId>azure-spring-boot</artifactId>
                 <version>${azure.spring.boot.version}</version>
             </dependency>
+            <dependency>
```
saragluna, line 122: No need to include this in a dependencyManagement here since we already got this covered in the jms starter

azure-spring-boot-parent/pom.xml
```diff
@@ -106,12 +105,7 @@
                 <artifactId>wiremock-standalone</artifactId>
                 <version>${wiremock-standalone.version}</version>
             </dependency>
-            <!--Qpid-->
```
saragluna, line 109: We should put the service-bus-jms dependency here.

PR #905, opened 2020-05-29
Keyvault performance

mnriem:
## Summary
Rework refreshing logic to increase performance

## Issue Type
- Bug fixing

## Starter Names

## Additional Information


azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
```diff
@@ -3,91 +3,148 @@
...
+
+    /**
+     * Stores the secret client.
+     */
+    private final SecretClient secretClient;
+
+    /**
+     * Stores the secret keys.
+     */
+    private List<String> secretKeys;
```
chenrujun, line 43: To make code less confusing, I suggest to change variable/method name:

* secretKeys -> configuredSecretKeys
* properties  -> keyVaultSecretCache
* refreshProperties -> refreshkeyVaultSecretCache

azure-spring-boot/src/test/java/com/microsoft/azure/keyvault/spring/CaseSensitiveKeyVaultTest.java
```diff
@@ -25,21 +23,18 @@
...
     @Test
     public void testGet() {
-        final List<String> keys = Arrays.asList("key1", "Key2");
-        
         final KeyVaultOperation keyVaultOperation = new KeyVaultOperation(
                 keyVaultClient,
-                "https:fake.vault.com",
-                TOKEN_ACQUIRE_TIMEOUT_SECS,
-                keys,
+                DEFAULT_REFRESH_INTERVAL_MS,
```
chenrujun, line 28: Please change `DEFAULT_REFRESH_INTERVAL_MS` to 1 minute.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
```diff
@@ -3,91 +3,148 @@
...
+            final Timer timer = new Timer();
+            final TimerTask task = new TimerTask() {
+                @Override
+                public void run() {
+                    refreshProperties();
+                }
+            };
+            timer.scheduleAtFixedRate(task, 0, refreshInMillis);
+        } else {
+            refreshProperties();
```
chenrujun, line 73: If `refreshInMillis <= 0`, we will not use cache, we should get the result(secret name list and secret value) from key-vault in every request.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
```diff
@@ -3,91 +3,148 @@
...
+        }
+    }
+
+    /**
+     * Get the property.
+     *
+     * @param property the property to get.
+     * @return the property value.
+     */
+    public String getProperty(String property) {
```
chenrujun, line 83: `getKeyVaultSecretValue` will be easier to understand.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
```diff
@@ -3,91 +3,148 @@
...
+
+        if (refreshInMillis > 0) {
+            final Timer timer = new Timer();
+            final TimerTask task = new TimerTask() {
+                @Override
+                public void run() {
+                    refreshProperties();
+                }
+            };
+            timer.scheduleAtFixedRate(task, 0, refreshInMillis);
```
chenrujun, line 71: If `refreshInMillis` equal to 1 second, and `task` will cost 1 minute to finish, then many task will run at the same time? That's not good.

chenrujun: And it's not necessary to `refreshProperties()` again and again if the app only need to read the properties once.

(cc @jialindai FYI.)

chenrujun: I mean, if `getPropertyNames()` and `getProperty` are not called, `refreshProperties()` should not execute.

chenrujun:
Hi, @mnriem , 
(cc @yiliuTo , FYI).

What's current process of this PR? 
If you are busy, maybe I can help to finish the logic of cache key vault values.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
```diff
@@ -3,91 +3,148 @@
...
-    private final SecretClient keyVaultClient;
-    private final String vaultUri;
-    private volatile List<String> secretNames;
-    private final boolean secretNamesAlreadyConfigured;
-    private final long secretNamesRefreshIntervalInMs;
-    private volatile long secretNamesLastUpdateTime;
+    /**
+     * Stores the properties.
+     */
+    private LinkedHashMap<String, String> properties = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
chenrujun: Thank you.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
chenrujun: OK.

azure-spring-boot/src/test/java/com/microsoft/azure/keyvault/spring/CaseSensitiveKeyVaultTest.java
chenrujun: But current value is 1800s (30 minutes), that's too long.
How about 3 minutes?

(cc @jialindai, @yiliuTo , what's your opinion?)

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
chenrujun: > Should we allow for no caching at all?

I already discussed with @jialindai , we think that we allow no caching at all. 

Let user configure whether need to cache.

If user configure `refreshInMillis == 0`, not cache make sense. 

Maybe can can use `kay-vault.cache.ms` instead of `refresh-interval` as configure item?

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
chenrujun: OK.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
chenrujun: > I disagree. A refresh interval is exactly that, aka the interval after which the keys/values are refreshed from the backend. So not going to change this. If the application developer does not need a refresh of the keys/values after startup the refresh interval can be set to 0.

If user configure refresh time to 5 minutes, and getPropertyNames() and getProperty are not called, refreshProperties() every 5 minutes is waste of resource, and will be a burden to key-vault service.

Maybe `cache-time` is better configure item than `refresh-interval`?

chenrujun: >From the JavaDoc of Timer, "Corresponding to each Timer object is a single background thread that is used to execute all of the timer's tasks, sequentially."

Then what happen if refresh-interval  = 1 minute and refresh operation need 2 minutes?

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
mnriem: No caching will be detrimental to any application as you are not in control of when the PropertySource gets consulted. I strongly advice against that.

azure-spring-boot/src/test/java/com/microsoft/azure/keyvault/spring/CaseSensitiveKeyVaultTest.java
mnriem: What is the use.case necessitating this huge change of the refresh interval?

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
mnriem: I feel you are looking to implement a TTL for each secret we would have to rearchitect this. Can we proceed as it currently stands? And IF we decide that TTL is appropriate to do that later?

mnriem: If the refresh takes longer than 2 minutes I would argue we have a problem with Azure KeyVault. But even with that said why on earth would anyone want to refresh the cache every minute? If there are doing that then they are using this integration for something other than its intended purpose and they would be better off using the SDK to interact with Azure KeyVault.

azure-spring-boot/src/test/java/com/microsoft/azure/keyvault/spring/CaseSensitiveKeyVaultTest.java
chenrujun: Out customer may update key-value pair in key-vault, then get the pair by `environment.getProperty()` minutes latter. if customer can not get the updated data, it may make our customer confused.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
chenrujun: Your concern make sense.

But current behavior is no cache. 

IMU,
* About 80%(just guess) of key-vault property will only get one time. (when start the app).
* It's OK to let our customer determine whether to use cache.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
chenrujun: > I feel you are looking to implement a TTL for each secret, if so we would have to rearchitect this.

Not for each secret, use current configure item `refresh-interval` as TTL of all secrets is my purpose.

> Can we proceed as it currently stands? And IF we decide that TTL is appropriate to do that later?

Can we use  current configure item `refresh-interval` as TTL of all secrets?

chenrujun: > But even with that said why on earth would anyone want to refresh the cache every minute?

Just as I said above: our customer may update key-value pair in key-vault (manually or by java code), then get the pair by environment.getProperty() minutes latter. if customer can not get the updated data, it may make our customer confused.

azure-spring-boot/src/test/java/com/microsoft/azure/keyvault/spring/CaseSensitiveKeyVaultTest.java
mnriem: I am inclined to change the refresh value to 0 and document that the PropertySource only accesses the KeyVault upon startup and IF the customer wants to use the @Environment to get new values out of the KeyVault they need to set it according to their requirements.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
mnriem: If we set the refresh interval to -1 to indicate we do NO caching at all then I will go ahead and implement that, but I still think this is bad from a user experience point of view. Our code now can slow down any part of their application because we do NOT control the order in which PropertySources are consulted, nor at what time! So agreed on the -1?

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
mnriem: > > I feel you are looking to implement a TTL for each secret, if so we would have to rearchitect this.
> 
> Not for each secret, use current configure item `refresh-interval` as TTL of all secrets is my purpose.
> 
> > Can we proceed as it currently stands? And IF we decide that TTL is appropriate to do that later?
> 
> Can we use current configure item `refresh-interval` as TTL of all secrets?

Great then I can leave the code I wrote as is as that is what the refresh interval does. It refreshes all the secrets after a specific time (Time To Live)

mnriem: > > But even with that said why on earth would anyone want to refresh the cache every minute?
> 
> Just as I said above: our customer may update key-value pair in key-vault (manually or by java code), then get the pair by environment.getProperty() minutes latter. if customer can not get the updated data, it may make our customer confused.

See my comments above

azure-spring-boot/src/test/java/com/microsoft/azure/keyvault/spring/CaseSensitiveKeyVaultTest.java
chenrujun: > I am inclined to change the refresh value to 0 and document that the PropertySource only accesses the KeyVault upon startup

Sounds good. I have the same idea before.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
chenrujun: As discussed in email. Let our customer to decide the TTL of cache.

azure-spring-boot/src/main/java/com/microsoft/azure/keyvault/spring/KeyVaultOperation.java
chenrujun: >  It refreshes all the secrets after a specific time (Time To Live)

To save the resource of network and key-vault server:
> I mean, if getPropertyNames() and getProperty are not called, refreshProperties() should not execute.

chenrujun:
@jialindai said that

```
Manfred had some discussion with us on PR 905, the conclusion is go ahead with Manfred’s current approach which eagerly pull key vault. Please help to approve it.

There may be some further fine tuning, which will be tracked by further PR.
```

So I'll approve it now.

AbhijeetAvaghade:
if this is part of 2.3.2 release?

yiliuTo:
> if this is part of 2.3.2 release?

We plan to release it in 2.3.3 version

AbhijeetAvaghade:
Any tentative date for 2.3.3 release?

yiliuTo:
> Any tentative date for 2.3.3 release?

We plan to release 2.3.3 in mid-August, around 8.11.

PR #929, opened 2020-07-31
Sort lines.  Add link to storage starter.

edburns:
Based on @jialindai 's comment on https://dev.azure.com/azure-spring-integration/spring-integration-private/_workitems/edit/254 we should add the spring boot storage starter.  I did not know we had one because I based my understanding on what are the available starters on this bullet list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,11 @@
 
 ## Below are the new locations for Spring modules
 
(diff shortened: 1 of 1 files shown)


saragluna:
Issue created https://github.com/Azure/azure-sdk-for-java/issues/13723.

